PR #2, opened 2023-02-14
Activate tests

benaadams:
@Scooletz disabled one test as it never seemed to complete; also one test is currently failing


Scooletz:
@benaadams I'm OOO at the moment and I have some local changes in preparation for contributions. I'll take a look at it in ~weeks. Currently the decision of whether to push with project is somewhat dependent on the EF Grant I issued for.

PR #39, opened 2023-04-14
ReadOnly batches

### LINKED ISSUE #5: Reader transactions
Introduce reader transactions that does not lock writer. This should be done in a manner similar to LMDB, where each reader tx remembers the id of the root that it's accessing. Then, whenever a decision is made whether or not a page can be reused, the list can be traversed to find the oldest and act accordingly

https://github.com/LMDB/lmdb/blob/3947014aed7ffe39a79991fa7fb5b234da47ad1a/libraries/liblmdb/mdb.c#L2401-L2417

Scooletz:
This PR introduces a notion of `IReadOnlyBatch` that should be used for read-only operations. It allows to capture a snapshot of Paprika at a given state (block) and keep it alive arbitrarily long. The pages used by this batch won't be touched by write batches until the readonly batch is alive (not disposed).

This PR also limits the number of writers to `1`. 

Resolves #5

### DIFF
--- a/src/Paprika.Tests/BasePageTests.cs
+++ b/src/Paprika.Tests/BasePageTests.cs
@@ -1,46 +1,14 @@
 ﻿using System.Runtime.InteropServices;
-using NUnit.Framework;
 using Paprika.Pages;
 
 namespace Paprika.Tests;
 
 public abstract class BasePageTests
 {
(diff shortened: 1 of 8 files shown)


Scooletz:
Got caught with an uninitialized memory 😅 It's green now.

Issue #34, opened 2023-04-03
Build in storage support for accounts


Scooletz:
The storage approach needs to be designed and implemented. It might require different kind of pages as framing for accounts will be much different. The sketched approach is to have a tuple of three `(UInt256 key,UInt256 value, PageAddress next)` and provide a fanout scenario similar to the one used for the accounts. The modified approach though, would be to allow the frames to reside on the same page as the account data. This would require even more fine-grained pooling (#23)

Please do update docs according to the design.

Scooletz:
We could use `Serializer.Account` introduced at #54 and build on top of that, so for small storages, they would be serialized alongside the contract. For bigger, the `FixedMap` push down could be used if the counting is a bit augmented. This would allow to have `DataPage`s that contain some accounts, some small contracts (small in a sense of the number of storage items) and only when the page is overflow for contracts, flush it down, potentially introducing a new kind of page `StoragePage` that would hold information only about the given account and provide the root for the storage trie.

PR #49, opened 2023-04-18
Multiple frames

### LINKED ISSUE #23: Better page packing
Provide capability to pack data into a page in a better way. Use the page approach where the slice of data responsible for storing kv pairs grows from two directions to the center:

1. id entries, grow ➡, from 0 to the center and capture basic lookup information
2. payload, grow ⬅ , from the end to the center and capture payload

Additional materials:

1. PostgreSQL: [bufpage.c](https://github.com/postgres/postgres/blob/26f36fe844880ff68dc28493bba9aced5aad0034/src/backend/storage/page/bufpage.c)
1. PostgresSQL [docs](https://www.postgresql.org/docs/current/storage-page-layout.html)
1. Andy Pavlo lecture: [03 - Database Storage 1 (CMU Intro to Database Systems / Fall 2022](https://youtu.be/df-l2PxUidI?t=3055)

This implementation might not necessarily deal with all the aspects, but should be similar to the slotted pages. The indexing aspect might left as is now.

Scooletz:
This PR introduces multiple [frame](https://github.com/NethermindEth/Paprika/blob/main/docs/design.md#frames) types to allow some space savings when storing `EOA` data.

It should include a preliminary work for #34 

Resolves #23


Scooletz:
Closing for now. Will be revived soon as a separate PR

PR #64, opened 2023-05-01
In page buffer

Scooletz:
This is a speculative experimental PR that uses in-page buffer to amortize the cost of IO. Previously a `DataPage` flushed one of its buckets, it was no longer storing keys with the given nibble. With this PR it does and flushes occasionally, whenever the in-page `FixedMap` is full. 

### Drawbacks

The drawback is that it will probably make the Merkle part terribly hard. Do not merge now.

### Benchmarks

Results:
- number of pages used by block: ~300 
    - 75% drop from previous 1857
    - this means effectively only ~1.2MB of writes per block
    - the number of pages per block does not change with the number of accounts getting bigger! remains constant. This is a big win as it's a good indicator for bigger data sets!
    - SSDs will last longer (1/4 of writes)
    - flushes are faster
- flush to disk: 0.11 s
    - ~50% drop from previous 0.2 s
- disk size: 4.17GB
    - 20% increase from previous 3.33GB
    - it's the least measurable as no assertion over how pages are filled, requires clarifying

### DIFF
--- a/src/Paprika.Tests/BasePageTests.cs
+++ b/src/Paprika.Tests/BasePageTests.cs
@@ -17,13 +17,13 @@ public abstract class BasePageTests
     {
         private readonly Dictionary<DbAddress, Page> _address2Page = new();
         private readonly Dictionary<UIntPtr, DbAddress> _page2Address = new();
+        private readonly HashSet<DbAddress> _written = new();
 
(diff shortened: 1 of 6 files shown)


Scooletz:
Closed in favor of modernized #82

PR #78, opened 2023-05-09
Add pipeline for benchmarks on linux machine

stdevMac:
This PR introduces a pipeline that build the recent version of `Paprika.Runner` and runs it in a benchmark environment. The output is reported to a file.

### DIFF
--- a/src/Paprika.Runner/Program.cs
+++ b/src/Paprika.Runner/Program.cs
@@ -23,7 +23,7 @@ public static class Program
 
     private const int NumberOfLogs = 10;
 
-    private const long DbFileSize = 10 * Gb;
+    private const long DbFileSize = 6 * Gb;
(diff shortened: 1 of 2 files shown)


src/Paprika.Runner/Program.cs
```diff
@@ -23,7 +23,7 @@ public static class Program
 
     private const int NumberOfLogs = 10;
 
-    private const long DbFileSize = 10 * Gb;
+    private const long DbFileSize = 6 * Gb;
```
LukaszRozmej, line 26: why not read it from arguments?

Scooletz: Will be squeezing more, the further we go. And yes, later, this will be automatically set by the db.

Scooletz (changes requested):

Scooletz:
Closed in favor of #136

Issue #77, opened 2023-05-09
Merkleization


Scooletz:
Implement Merkleization for the tree

- [x] precommit hook for injecting Merkleization  #96  
- [x] serialization of `Branch`, `Leaf` and `Extension` to a `Span` #98 
- [x] calculation of Keccaks for the nodes  #104 
- [x] capturing all keys inserted in a block and passing them as the precommit to the Merkleization #111
- [x] constructing the Merkle root hash by walking the keys from the above #113
- [x] tracking the dirty status for Merkle tree nodes #113
- [x] deletes #113 
- [ ] Keccak memoization
- [ ] storage handling (separate storage in block and bottom up calculation)
- [ ] squashing into `PagedDb`

The last 3 will be addressed in separate issues.

emlautarom1:
To start working on this, I suggest writing some test on the current Nethermind client (here, probably: https://github.com/NethermindEth/nethermind/blob/22cc2b238e72e59a28c4b9e99019a4a145cc5895/src/Nethermind/Nethermind.Trie.Test/TrieTests.cs#L26) and check what we get as the `RootHash`. Then, migrate those tests into Paprika and use the obtained hashes as expected values during testing.

emlautarom1:
Following the approach described above, I've written the following test in the Nethermind client (NC moving forward) and translated it into Paprika:

```C#
[Test]
public void Single_account()
{
    MemDb memDb = new();
    using TrieStore trieStore = new(memDb, Prune.WhenCacheReaches(1.MB()), Persist.EveryBlock, _logManager);
    PatriciaTree patriciaTree = new(trieStore, _logManager);

    var key = new Keccak(new byte[] { 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7 });
    var account = new Account(23, 13);

    patriciaTree.Set(key.Bytes, new AccountDecoder().Encode(account).Bytes);
    patriciaTree.Commit(blockNumber: 0);
    patriciaTree.UpdateRootHash();

    var rootHash = patriciaTree.RootHash;
    TestContext.WriteLine(rootHash);
}
```

This, translated to Paprika looks something like:

```C#
[Test]
public void Single_account()
{
    using var db = PagedDb.NativeMemoryDb(SmallDb);

    var key = Values.Key0;
[...]

PR #89, opened 2023-05-30
Add support for Merkleization

emlautarom1:
Implements #77


src/Paprika.Tests/MerkleTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        using var db = PagedDb.NativeMemoryDb(SmallDb);
+
+        using (var batch = db.BeginNextBatch())
+        {
+            batch.Commit(CommitOptions.FlushDataAndRoot);
+        }
+
+        using (var batch = db.BeginReadOnlyBatch())
+        {
+            var rootHash = batch.GetRootHash();
```
Scooletz, line 23: I think we should leave the db agnostic in regards to the calculations. As implemented right now in #88 , there is a clear split between blocks that are `latest` or `safe` and blocks that are `finalized`. Only the latter will be stored in the database in a form that makes it squashed together. If you build up the tests maybe it would be useful to build it on level, where you imperatively state `Set for leaves values'. After all, when the left-fold is applied, all the information will be squashed together.

src/Paprika/Tree/Leaf.cs
```diff
@@ -0,0 +1,65 @@
+using Paprika.Crypto;
+using Paprika.Data;
+using Paprika.RLP;
+
+namespace Paprika.Tree;
+
+public static class Leaf
```
Scooletz, line 7: No tests for it right now? 👁️

src/Paprika/Tree/Leaf.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    private static RlpStream RlpOfAccount(Account account, Keccak codeHash, Keccak storageRootHash)
+    {
+        var contentLength =
+            Rlp.LengthOf(account.Balance)
+            + Rlp.LengthOf(account.Balance)
+            + Rlp.LengthOfKeccakRlp // CodeHash
+            + Rlp.LengthOfKeccakRlp; // StorageRootHash
+
+        Span<byte> rlpBuffer = new byte[Rlp.LengthOfSequence(contentLength)];
```
Scooletz, line 36: Easy to optimize away later on 😃

Scooletz (commented):
I'm trying my best to merge #88 as soon as possible. Once it's there, it should be easier to discuss the finalized data set and more.

src/Paprika.Tests/MerkleKeccakTests.cs
```diff
@@ -0,0 +1,53 @@
...
+using Paprika.Tree;
+
+namespace Paprika.Tests;
+
+public class MerkleKeccakTests
+{
+    [Test]
+    [TestCase(0, "E2533A0A0C4F1DDB72FEB7BFAAD12A83853447DEAAB6F28FA5C443DD2D37C3FB")]
+    [TestCase(1, "DD358AF6B1D8E875FBA0E585710D054F14DD9D06FA3C8C5F2BAF66F413178F82")]
+    [TestCase(2, "A654F039A5F9E9F30C89F21555C92F1CB1E739AF11A9E9B12693DEDC6E76F628")]
```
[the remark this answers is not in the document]
emlautarom1: That's quite nice. I really disliked using indexes to deal with this.

src/Paprika/Tree/Leaf.cs
emlautarom1: We have 100% coverage of this class

src/Paprika/Tree/KeccakOrRlp.cs
```diff
@@ -0,0 +1,38 @@
...
+    public static KeccakOrRlp WrapRlp(Span<byte> data)
+    {
+        var destination = new byte[32];
+
+        if (data.Length < 32)
+        {
+            destination[0] = (byte)data.Length;
+            data.CopyTo(destination[1..]);
+            return new KeccakOrRlp(Type.Rlp, destination);
+        }
```
emlautarom1, line 31: We have no test case that covers this branch. It seems that for Leaves it is impossible, but it's worth mentioning. I initially made a mistake of doing `return new KeccakOrRlp(Type.Rlp, data);` instead of `return new KeccakOrRlp(Type.Rlp, destination);`

src/Paprika.Tests/MerkleTests.cs
emlautarom1: I'm not sure I quite follow. You mean separating `computing/setting the hashes` from `get the root hash`? Could you rephrase the last two sentences?

Scooletz: I think we could have the batch a bit more agnostic about the `RootHash`. Yes, in #88 there's the metadata, but maybe we could leave it and just operate on the tree with setting `Set(Key.KeccakOrRlp(NibblePath.Empty), ...` ? We can leave as is for now and discuss further. It does not break the idea that much. Let's pause this discussion till #88 lands.

src/Paprika/Tree/KeccakOrRlp.cs
Scooletz: It could be the case if a path is 1 nibble only and the account is almost empty 👀 , like 0.0000....1 Eth and 0 nonce?

src/Paprika/Tree/Leaf.cs
```diff
@@ -0,0 +1,65 @@
...
+        var result = EncodeLeaf(nibblePath, rlpStream.Data);
+
+        return result;
+    }
+
+    private static RlpStream RlpOfAccount(Account account, Keccak codeHash, Keccak storageRootHash)
+    {
+        var contentLength =
+            Rlp.LengthOf(account.Balance)
+            + Rlp.LengthOf(account.Balance)
```
emlautarom1, line 32: This should be ` Rlp.LengthOf(account.Balance) +  Rlp.LengthOf(account.Nonce)`

emlautarom1: Fixed. Also added test cases to trigger this bug

src/Paprika/Merkle/Node.cs
```diff
@@ -0,0 +1,81 @@
...
+
+public enum NodeType : byte
+{
+    Leaf,
+    Extension,
+    Branch
+}
+
+[StructLayout(LayoutKind.Explicit, Pack = 1, Size = Size)]
+public readonly struct Branch
```
emlautarom1, line 15: A Branch is 35 bytes long if we always want to keep the Keccak (which we initially do):
- 1 byte for `Header`, which contains `Dirty` and `NodeType`
- 2 bytes for `Nibbles`, a Bit Set of nibbles
- 32 bytes for `Keccak` (initially, always present)

src/Paprika/Merkle/Node.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        var mask = 1;
+        for (byte i = 0; i < 16; i++, mask <<= 1)
+        {
+            if ((_nibbles & mask) != 0)
+            {
+                yield return i;
+            }
+        }
+    }
```
emlautarom1, line 44: Initial approach on enumerating the nibbles from the BitSet

src/Paprika/Merkle/Node.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    // Extension: 1 byte for type + dirty, var-length for path (usually it will be a nibble or a few)
+    private const int Size = 33;
+    private const byte IsDirtyFlag = 0b1000;
+    private const byte NodeTypeFlag = 0b0110;
+
+    [FieldOffset(0)]
+    private readonly byte _header;
+
+    public NibblePath Path => default;
```
emlautarom1, line 67: I'm not quite sure how to store a var-length `NibblePath`

src/Paprika/Merkle/Node.cs
```diff
@@ -0,0 +1,81 @@
...
+    // TODO: Should we expose ushort? What about a Span of two bytes?
+    public Branch(ushort nibbles)
+    {
+        _header = (byte)NodeType.Branch << 1 | IsDirtyFlag;
+        _nibbles = nibbles;
+    }
+}
+
+[StructLayout(LayoutKind.Explicit, Pack = 1, Size = Size)]
+public readonly struct Extension
```
emlautarom1, line 57: The size of an `Extension` depends on the size of the `NibblePath`.
- 1 byte for Header, which contains Dirty and NodeType
- 1 byte for the `NibblePath` length (N)
- N bytes for the `NibblePath`

If `M` is the maximum size of a NibblePath, then the maximum size of an `Extension` is M + 2.

Note that there is no Keccak here

src/Paprika/Merkle/Node.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    public Extension(NibblePath path)
+    {
+        _header = (byte)NodeType.Extension << 1 | IsDirtyFlag;
+    }
+}
+
+[StructLayout(LayoutKind.Explicit, Pack = 1, Size = Size)]
+// Temporary name due to clash with existing `Leaf`
+public readonly struct LeafNode
```
emlautarom1, line 80: The size of a Leaf depends on the length of the NibblePath.

- 1 byte for Header, which contains Dirty and NodeType
- 32 bytes for Keccak
- 1 byte for the NibblePath length (N)
- N bytes for the NibblePath

If M is the maximum size of a NibblePath, then the maximum size of a Leaf is M + 34.

emlautarom1:
The PR will be split up in multiple separate PRs with smaller scope (see #98 for starters)

PR #96, opened 2023-06-16
Merkle foundations with pre-commit hook

Scooletz:
This PR introduces a seam for the place where `Merkle` construct will be created. It is done in a way, so that `Verkle` construct in a future should be also easy to inject into it as a similar, but separate, implementation. The implementation in the `blockchain` is not done yet, but it should not prevent users from using the seam.

Effectively, this allows to implement the Merkle construct as 

```
class Merkle : IPreCommitBehavior
{
    public void BeforeCommit(ICommit commit)
    {
      // build the Merkle construct and remember to store the root under `Key.Merkle('')`
    }
}

### DIFF
--- a/src/Paprika/Chain/Blockchain.cs
+++ b/src/Paprika/Chain/Blockchain.cs
@@ -44,6 +44,7 @@ public class Blockchain : IAsyncDisposable
     private readonly MetricsExtensions.IAtomicIntGauge _flusherQueueCount;
 
     private readonly PagedDb _db;
+    private readonly IPreCommitBehavior? _preCommit;
     private readonly TimeSpan _minFlushDelay;
     private readonly Action? _beforeMetricsDisposed;
     private readonly Task _flusher;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/Paprika/Chain/Blockchain.cs [hunk 1]
```diff
@@ -417,6 +424,8 @@ private void Set(in Key key, in ReadOnlySpan<byte> payload)
             map.TrySet(key, payload);
         }
 
+        public IKeyEnumerator GetEnumerator() => throw new NotImplementedException("Not implemented yet");
```
emlautarom1, line 427: If `Block` is responsible for keeping track of the keys used, could we face performance issues?

src/Paprika/Chain/Blockchain.cs
```diff
@@ -391,7 +396,9 @@ public void SetStorage(in Keccak key, in Keccak address, UInt256 value)
             Set(Key.StorageCell(path, address), payload);
         }
 
-        private void Set(in Key key, in ReadOnlySpan<byte> payload)
+        public bool TryGet(in Key key, out ReadOnlySpanOwner<byte> result) => throw new NotImplementedException("Not implemented yet");
```
emlautarom1, line 399: We already have a `Get` method, why do we need a `TryGet`?

emlautarom1 (approved):
Interfaces look good, I was thinking something similar to this.

src/Paprika/Chain/Blockchain.cs
Scooletz: This is a subject to debate what to do on the missing one. I assume for now that you'd like to know that it was not there 👀

Scooletz:
Merge at will @emlautarom1

PR #97, opened 2023-06-21
Fix for the write error

Scooletz:
This PR focuses on fixing the overflow error 😭 . The data were written at the wrong page as the offset silently overflown in the `MemoryMappedPageManager`. Now, with a new property of `DbAddress` that does the calculation in one place this should not be the case.

Additionally, a few `[DebuggerStepThrough]` are added to help future battles.

One more thing is that `AwaitWrites` are properly awaited at the end of the writing loop!

### DIFF
--- a/src/Paprika.Runner/Program.cs
+++ b/src/Paprika.Runner/Program.cs
@@ -14,7 +14,7 @@ namespace Paprika.Runner;
 
 public static class Program
 {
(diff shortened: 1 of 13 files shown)


emlautarom1 (approved):
LGTM.

As discussed, it would be nice to have checked arithmetic during debug builds to avoid this kind of bugs in the future.

Scooletz:
This is hidden as a property. Let me add one test for this then!

PR #104, opened 2023-06-23
Merkle Node Keccak

emlautarom1:
Compute the `Keccak` or `RLP` of Merkle Nodes (Branches, Leaves and Extensions)

### DIFF
--- a/src/Paprika.Tests/Merkle/HashingTests.cs
+++ b/src/Paprika.Tests/Merkle/HashingTests.cs
@@ -43,6 +43,6 @@ public class HashingTests
         var expectedHash = new Keccak(Convert.FromHexString(hexString));
 
         Assert.That(computedHash.DataType, Is.EqualTo(KeccakOrRlp.Type.Keccak));
-        Assert.That(new Keccak(computedHash.Data), Is.EqualTo(expectedHash));
+        Assert.That(new Keccak(computedHash.AsSpan()), Is.EqualTo(expectedHash));
     }
 }
--- a/src/Paprika.Tests/Merkle/HashingTests.cs
+++ b/src/Paprika.Tests/Merkle/HashingTests.cs
@@ -1,0 +1,48 @@
+using Nethermind.Int256;
+using NUnit.Framework;
+using Paprika.Crypto;
+using Paprika.Data;
+using Paprika.Merkle;
+using Paprika.RLP;
+
+namespace Paprika.Tests.Merkle;
+
+public class HashingTests
+{
+    private static object[] _keysBalancesNoncesHexStrings =
+    {
+        new object[]
+        {
+            Values.Key0, Values.Balance0, Values.Nonce0,
+            "E2533A0A0C4F1DDB72FEB7BFAAD12A83853447DEAAB6F28FA5C443DD2D37C3FB"
+        },
+        new object[]
+        {
+            Values.Key1A, Values.Balance1, Values.Nonce1,
+            "DD358AF6B1D8E875FBA0E585710D054F14DD9D06FA3C8C5F2BAF66F413178F82"
+        },
+        new object[]
+        {
+            Values.Key2, Values.Balance2, Values.Nonce2,
+            "A654F039A5F9E9F30C89F21555C92F1CB1E739AF11A9E9B12693DEDC6E76F628"
+        },
+        new object[]
+        {
+            Values.Key0, UInt256.MaxValue, Values.Nonce0,
+            "FC790A3674B6847C609BB91B7014D67B2E70FF0E60F53AC49A8A45F1FECF35A6"
+        },
+    };
+
+    [Test]
+    [TestCaseSource(nameof(_keysBalancesNoncesHexStrings))]
+    public void Leaf_keccak(Keccak key, UInt256 balance, UInt256 nonce, string hexString)
+    {
+        var account = new Account(balance, nonce);
+
+        var computedHash = Node.Leaf.KeccakOrRlp(NibblePath.FromKey(key), account);
+        var expectedHash = new Keccak(Convert.FromHexString(hexString));
+
+        Assert.That(computedHash.DataType, Is.EqualTo(KeccakOrRlp.Type.Keccak));
+        Assert.That(new Keccak(computedHash.Data), Is.EqualTo(expectedHash));
+    }
+}
--- a/src/Paprika.Tests/Merkle/HashingTests.cs
+++ b/src/Paprika.Tests/Merkle/HashingTests.cs
@@ -39,7 +39,7 @@ public class HashingTests
     {
         var account = new Account(balance, nonce);
 
-        var computedHash = Node.Leaf.KeccakOrRlp(NibblePath.FromKey(key), account);
+        Node.Leaf.KeccakOrRlp(NibblePath.FromKey(key), account, out var computedHash);
         var expectedHash = new Keccak(Convert.FromHexString(hexString));
 
         Assert.That(computedHash.DataType, Is.EqualTo(KeccakOrRlp.Type.Keccak));
--- a/src/Paprika.Tests/Merkle/HashingTests.cs
+++ b/src/Paprika.Tests/Merkle/HashingTests.cs
@@ -9,27 +9,34 @@ namespace Paprika.Tests.Merkle;
 
 public class HashingTests
 {
+    // NOTE: Expected results come from:
+    // https://github.com/NethermindEth/nethermind/blob/bb7e792722085294b5adea54b598f3efd5887689/src/Nethermind/Nethermind.Trie.Test/PaprikaTrieTests.cs
     private static object[] _keysBalancesNoncesHexStrings =
     {
         new object[]
         {
             Values.Key0, Values.Balance0, Values.Nonce0,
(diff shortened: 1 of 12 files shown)


src/Paprika.Tests/Merkle/RootHashTests.cs
```diff
@@ -0,0 +1,52 @@
...
+    private const int SmallDb = 256 * Page.PageSize;
+    private readonly Keccak _blockKeccak = Keccak.Compute("block"u8);
+
+    [Test]
+    public async Task Empty_database()
+    {
+        using var db = PagedDb.NativeMemoryDb(SmallDb);
+        await using var blockchain = new Blockchain(db, preCommit: new ComputeMerkleBehavior());
+
+        using var block = blockchain.StartNew(Keccak.Zero, _blockKeccak, 1);
```
emlautarom1, line 20: It would be nice if we could extract this into a `Setup` method

src/Paprika/Merkle/Hashing.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        private static RlpStream RlpOfAccount(Account account, Keccak codeHash, Keccak storageRootHash)
+        {
+            var contentLength =
+                Rlp.LengthOf(account.Balance)
+                + Rlp.LengthOf(account.Nonce)
+                + Rlp.LengthOfKeccakRlp // CodeHash
+                + Rlp.LengthOfKeccakRlp; // StorageRootHash
+
+            Span<byte> rlpBuffer = new byte[Rlp.LengthOfSequence(contentLength)];
```
emlautarom1, line 38: Let's avoid allocations using `stackalloc`.

src/Paprika/Merkle/Hashing.cs
```diff
@@ -0,0 +1,68 @@
...
+            rlpStream.Encode(account.Balance);
+            rlpStream.Encode(storageRootHash);
+            rlpStream.Encode(codeHash);
+
+            return rlpStream;
+        }
+
+        private static KeccakOrRlp Encode(NibblePath path, Span<byte> value)
+        {
+            Span<byte> hexPath = new byte[path.HexEncodedLength];
```
emlautarom1, line 52: Let's avoid allocations using `stackalloc`.

src/Paprika/Merkle/Hashing.cs
```diff
@@ -0,0 +1,68 @@
...
+            path.HexEncode(hexPath, true);
+
+            var contentLength = Rlp.LengthOf(hexPath) + Rlp.LengthOf(value);
+            var totalLength = Rlp.LengthOfSequence(contentLength);
+
+            Span<byte> data = new byte[totalLength];
```
emlautarom1, line 58: Let's avoid allocations using `stackalloc`.

src/Paprika/RLP/Rlp.cs
```diff
@@ -0,0 +1,65 @@
+using Nethermind.Int256;
+using Paprika.Utils;
+
+namespace Paprika.RLP;
+
+public static class Rlp
```
emlautarom1, line 6: There are multiple "magic numbers" in the class. We should extract them into constants and document the reason for their existence (spec, optimization, etc.)

src/Paprika/RLP/RlpStream.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    public void Encode(in Keccak keccak)
+    {
+        // TODO: If keccak is a known one like `Keccak.OfAnEmptyString` or `Keccak.OfAnEmptySequenceRlp`
+        // we can cache those `Rlp`s to be reused
+
+        WriteByte(160);
+        Write(keccak.BytesAsSpan);
```
emlautarom1, line 102: Optimization opportunity by reusing results on known inputs.

src/Paprika.Tests/Merkle/HashingTests.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Paprika.Tests.Merkle;
+
+public class HashingTests
+{
+    private static object[] _keysBalancesNoncesHexStrings =
+    {
+        new object[]
+        {
+            Values.Key0, Values.Balance0, Values.Nonce0,
```
Scooletz, line 16: Would be super useful if you could provide a link to a branch in NethermindEth maybe where you calculate these. Once they are merged, it will be fine but for now, keeping any kind of link would make the life much easier.

src/Paprika.Tests/Merkle/RootHashTests.cs
Scooletz: Is it needed though? My thinking was that `IPreCommitBehavior` is sufficient to work with, at least with the calculation? It can be called with an arbitrary `ICommit` mocked implementation.

src/Paprika/Merkle/Hashing.cs
Scooletz: `stackalloc byte[Rlp.LengthOfSequence(contentLength)]` it is, isn't it? 👁️

emlautarom1: We cannot do that since the `rlpBuffer` is wrapped into a `RlpStream` which we return at the end of the function.

src/Paprika.Tests/Merkle/RootHashTests.cs
emlautarom1: The tests insert multiple accounts, then commit the changes in order to trigger the Merkle `IPreCommitBehavior` to recompute the root hash. For that, we actually need to store the account (and all intermediate Merkle nodes) somewhere, and for that we need this setup code.

I don't quite follow how we can get away with just a mocked `ICommit`. Could you elaborate a little?

emlautarom1:
Removed all allocations with a little help of `scoped` and manually inlining methods.

src/Paprika/RLP/KeccakOrRlp.cs
```diff
@@ -0,0 +1,48 @@
+using Paprika.Crypto;
+
+namespace Paprika.RLP;
+
+public readonly ref struct KeccakOrRlp
```
emlautarom1, line 5: I also tried a different approach that uses a fixed size array instead of a `Keccak`:

```c#
public ref struct KeccakOrRlp
{
    public enum Type : byte
    {
        Keccak = 0,
        Rlp = 1,
    }

    private unsafe ref struct Buffer
    {
        private const int Size = 32;

#pragma warning disable CS0649
        // Compiler cannot figure out that the fields is being used
        private fixed byte _data[Size];
#pragma warning restore CS0649

        public Span<byte> AsSpan()
        {
            fixed (byte* p = _data)
            {
                return new Span<byte>(p, Size);
            }
        }
    }

    public Type DataType { get; }
    private Buffer _buffer;

    public Span<byte> AsSpan() => _buffer.AsSpan();

    private KeccakOrRlp(Type dataType, Buffer buffer)
    {
        DataType = dataType;
        _buffer = buffer;
    }

    public static KeccakOrRlp FromSpan(scoped Span<byte> data)
    {
        var buffer = new Buffer();
        var output = buffer.AsSpan();

        if (data.Length < 32)
        {
            output[0] = (byte)data.Length;
            data.CopyTo(output[1..]);
            return new KeccakOrRlp(Type.Rlp, buffer);
        }
        else
        {
            KeccakHash.ComputeHash(data, output);
            return new KeccakOrRlp(Type.Keccak, buffer);
        }
    }
}
```

On benchmarks, this approach is as performant as the `Keccak` implementation, while in my opinion being a bit clearer since we're not using a `Keccak` as a backing field (which, when we store an `Rlp` is not even a `Keccak`).

I'm not fully sure though if we're performing additional copies though, but I don't see any on the IL (besides the required ones).

src/Paprika.Tests/Merkle/HashingTests.cs
emlautarom1: I've created a branch with the same tests: https://github.com/NethermindEth/nethermind/tree/feature/paprika_merkle_tests.

src/Paprika.Tests/Merkle/RootHashTests.cs
emlautarom1: For now I've decided to keep it as is: the cost of avoiding the repetition is higher than repeating the same 3 lines of code. If this becomes an issue, we can revisit it.

src/Paprika/RLP/KeccakOrRlp.cs
Scooletz: This should work as you describe.

src/Paprika.Tests/Merkle/HashingTests.cs
Scooletz: Could you commit this link to the repository? Otherwise it will get lost. Just add it to the tests.

Scooletz (changes requested):
One remark and good to merge.

src/Paprika.Tests/Merkle/HashingTests.cs
emlautarom1: I actually did that but I forgot to push the changes. Whoops!

src/Paprika/RLP/KeccakOrRlp.cs
emlautarom1: Should we use this approach then? Or would you rather stick with the Keccak approach?

Scooletz (approved):
Good to go!

Issue #105, opened 2023-06-26
Optimize Account format


Scooletz:
Currently, when serializing `Account`, the `Nonce` and `Balance` store their lengths using full byte. Taking into consideration total Eth supply and how often nonce is bumped, both values should be capped at `UInt128` or less. Provide a way of storing accounts in an optimized way, where a single byte is used to encode both lengths:

- `0b1000_0000` - mask for marking the special encoding
- `0b0111_0000` - mask for storing the number of bytes for nonce (up to 7 bytes)
- `0b0000_1111` - mask for storing the number of bytes for balance (up to 15 bytes)

This should limit the account size by 1, which is ~2.5% of the total size and can greatly reduce the tree size.

CodeBlocker52:
To store the lengths of the Nonce and Balance fields in an optimized way using a single byte, you can follow the encoding scheme you provided. Here's a breakdown of the encoding scheme:

The first bit (0b1000_0000) is used as a mask to mark the special encoding.
The next four bits (0b0111_0000) are used to store the number of bytes used to encode the Nonce field.
The last four bits (0b0000_1111) are used to store the number of bytes used to encode the Balance field.
Using this encoding scheme, you can store both lengths efficiently in a single byte.
[Nethermind-Paprika-Code.zip](https://github.com/NethermindEth/Paprika/files/11929022/Nethermind-Paprika-Code.zip)

Below attached  the solution code for how you can implement the encoding and decoding functions in Python:

Scooletz:
@CodeBlocker52 For come reason, I missed this comment! Sorry!

PR #110, opened 2023-06-26
Account serialization optimization

Scooletz:
This PR makes the change in the way how accounts are serialized. For 99.9%, if not all, it should save 1 byte per account. As accounts are encoded usually within 60-80, it should save 1%, maybe 2% 👀 

Just to compare tree sizes with the `Paprika.Runner`:
- before: `Size of this Paprika tree: 337MB`
- after:  `Size of this Paprika tree: 319MB`

Solves #105


Scooletz:
Messed up git. Closing.

PR #151, opened 2023-08-29
Faster blocks by introducing a pooled dictionary

Scooletz:
This PR is a result of profiling block processing with a full blown Merkle construct. It was found that due to the nature of `SlottedArray` and its use in blocks, the performance was terrible as it was querying several pages multiple times. This resulted in long runs of the benchmarks. To fix that `PooledSpanDictionary` was introduced.

`PooledSpanDictionary` is a component that uses pages from `BufferPool` to store payloads of spans. To actually map values, `PooledSpanDictionary` hacks the equality comparer that allows to keep the reference local, without passing it in every single key and value. This keeps the keys and values small (structs of size of 8) while leaving the payload management to the pooled memory. When a block is disposed, all the pooled dictionaries return their pages to the pool which should keep it small.

### Before

image

### After

image


src/Paprika/Chain/PooledSpanDictionary.cs
```diff
@@ -0,0 +1,220 @@
...
+/// <remarks>
+/// This component hacks into the dictionary to provide the underlying keys by implementing the custom equality comparer.
+/// It allows to store in the underlying dict just simple entries and not pass the reference in there. To get the reference,
+/// the <see cref="Equals"/> gets the underlying <see cref="Span{Byte}"/> from pages;
+/// </remarks>
+public class PooledSpanDictionary : IEqualityComparer<PooledSpanDictionary.KeySpan>, IDisposable
+{
+    private const int BufferSize = BufferPool.BufferSize;
+    private readonly BufferPool _pool;
+    private readonly Dictionary<KeySpan, ValueSpan> _dict;
```
LukaszRozmej, line 19: Should this dictionary ever be cleared?

src/Paprika/Chain/PooledSpanDictionary.cs
```diff
@@ -0,0 +1,220 @@
...
+    private readonly List<Page> _pages = new();
+
+    private Page _current;
+    private int _position;
+
+    private readonly byte[] _key = new byte[1024];
```
LukaszRozmej, line 25: Why 1024? Are we sure all keys and values will fit into that memory? And for keys only seems excessive? Shouldn't be `MaxKeySize`?

src/Paprika/Chain/PooledSpanDictionary.cs
Scooletz: It's dropped and not reused. If we consider pooling of the dict, then yes. For now, not needed.

src/Paprika/Chain/PooledSpanDictionary.cs
Scooletz: It's a dummy constant. Will provide better.

PR #220, opened 2024-01-24
Boundary node for snap

Scooletz:
This PR allows to set an artificial node representing a `Boundary` of a snap sync that later can be overwritten.

### DIFF
--- a/src/Paprika/Chain/SnapSync.cs
+++ b/src/Paprika/Chain/SnapSync.cs
@@ -1,0 +1,51 @@
+﻿using Paprika.Crypto;
+using Paprika.Data;
+using Paprika.Merkle;
+
+namespace Paprika.Chain;
+
+public static class SnapSync
+{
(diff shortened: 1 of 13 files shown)


src/Paprika/Chain/SnapSync.cs
```diff
@@ -0,0 +1,51 @@
...
+        payload[0] = Byte0;
+        payload[1] = Byte1;
+        value.Span.CopyTo(payload[PreambleLength..]);
+
+        return payload.Slice(0, BoundaryValueSize);
+    }
+
+    public static bool CanBeBoundaryLeaf(in Node.Leaf leaf)
+    {
+        return leaf.Path.HasOnlyZeroes();
```
damian-orzechowski, line 25: When adding a boundary node (path-keccak pair), the path needs to filled up to full width with zeros?

Scooletz: Nope. It's takes care of by the `SnapSync` part. Just use methods on the raw.

PR #250, opened 2024-02-23
ReadOnlyBatch caches ids

Scooletz:
This PR introduces a cache of ids for the `ReadOnlyBatch`. As the readonly batch is used to access data that are not in memory, caching ids may speed up each query against storage (Merkle or value) by a few lookups. This, if a given account has its storage Merkleized, should benefit a lot.

### DIFF
--- a/src/Paprika/Store/PagedDb.cs
+++ b/src/Paprika/Store/PagedDb.cs
@@ -357,7 +357,9 @@ public class PagedDb : IPageResolver, IDb, IDisposable
 
     private class ReadOnlyBatch(PagedDb db, RootPage root, string name) : IReportingReadOnlyBatch, IReadOnlyBatchContext
     {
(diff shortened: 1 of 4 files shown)


src/Paprika/Store/PagedDb.cs
```diff
@@ -356,6 +357,9 @@ private DbAddress SetNewRoot(RootPage root)
 
     private class ReadOnlyBatch(PagedDb db, RootPage root, string name) : IReportingReadOnlyBatch, IReadOnlyBatchContext
     {
+        private readonly ConcurrentDictionary<Keccak, uint> _idCache = new(Environment.ProcessorCount,
```
benaadams, line 360: Use non-blocking ConcurrentDictionary? https://github.com/VSadov/NonBlocking

PR #255, opened 2024-02-24
Use NonBlocking ConcurrentDictionary

benaadams:


src/Paprika/Store/PagedDb.cs
```diff
@@ -357,7 +357,10 @@ private DbAddress SetNewRoot(RootPage root)
 
     private class ReadOnlyBatch(PagedDb db, RootPage root, string name) : IReportingReadOnlyBatch, IReadOnlyBatchContext
     {
-        private readonly ConcurrentDictionary<Keccak, uint> _idCache = new(Environment.ProcessorCount,
+        [ThreadStatic]
+        private static ConcurrentDictionary<Keccak, uint>? s_cache;
```
Scooletz, line 361: I was thinking about caching it. Someday 😝

src/Paprika/Chain/Blockchain.cs
```diff
@@ -393,10 +393,12 @@ public void Finalize(Keccak keccak)
     /// </summary>
     private class BlockState : RefCountingDisposable, IWorldState, ICommit, IProvideDescription, IStateStats
     {
+        [ThreadStatic]
+        private static HashSet<ulong>? s_bloomCache;
```
Scooletz, line 397: Have you measured it? It could save a lot of memory as usually a single BlockState will be out there.
At the same time, maybe some Resettable should be used. `HashSet` do not trim on `Clear`.

Scooletz (approved):
Good to go! Some remarks only.

src/Paprika/Store/PagedDb.cs
```diff
@@ -370,14 +373,28 @@ public void Dispose()
...
             _disposed = true;
             db.DisposeReadOnlyBatch(this);
+
+            ReturnCacheToPool();
+
+            void ReturnCacheToPool()
+            {
+                var idCache = _idCache;
+                _idCache = null!;
+                ref var cache = ref s_cache;
```
Scooletz, line 383: @benaadams Should `s_cache` access be `volatile`? No barrier on return.

benaadams: Is threadstatic so doesn't matter too much; I take a `ref` of `s_cache` because accessing (finding) a threadstatic is more expensive than regular field, so once have its memory location work with that.

I use `Interlocked.Exchange` to get; not because of thead worry issues, but is single fast instruction whereas it would be multi and messier with taking a `ref` (for this reason) geting value, conditionally blanking  etc

src/Paprika/Store/PagedDb.cs
benaadams: 225 sec trace

Before 126GB of ConcurrentDictionaries allocated from TxPool

image

After 0 ConcurrentDictionaries allocated over same time period

image

src/Paprika/Chain/Blockchain.cs
benaadams: Yep, finding changes with a profiler sitting on maintnet 😉

HashSet was always allocating a lot in AddWithResize; as it starts at 0 elements. However if started with a high capacity then it would always allocate a large amount, so this seems a fair comprimize.

This HashSet is in BlockProcessing, ConcurrentDictionary allocations were in TxPool

PR #259, opened 2024-02-25
Use pooled array for fp in Xor filter

shubham-sonthalia:


src/Paprika/Utils/Xor8.cs
```diff
@@ -147,7 +147,7 @@ public Xor8(IReadOnlyCollection<ulong> keys)
 
         _seed = seed;
 
-        var fp = new byte[arrayLength];
+        IntPtr fp = Marshal.AllocHGlobal(arrayLength);
```
benaadams, line 150: Better to use `ArrayPool`? Here the allocations are still happening just unmanaged which is a harder heap to clean from fragmentation (making allocations more expesnive and memory use increase over time)

shubham-sonthalia: Yeah, makes sense. ArrayPool will give contiguous memory allocation, and that in turn will be more efficient as compared to unmanaged buffer. Thanks for pointing that out.

shubham-sonthalia:
Pls check now.

benaadams (approved):

Scooletz (changes requested):
This removes one allocation leaving `_fingerprints` in the managed world, but it's always one alloc less :-) 
Please run `dotnet format whitespace` on solution, otherwise builds will break.

shubham-sonthalia:
Done @Scooletz

Scooletz (approved):
🌶️

PR #262, opened 2024-02-27
Destroy storage content for a destroyed account

### LINKED ISSUE #179: Garbage Collect destroyed accounts

shubham-sonthalia:
Resolves #179


src/Paprika/Store/Page.cs
```diff
@@ -29,6 +29,8 @@ public interface IPageWithData<TPage> : IPage
     Page Set(in NibblePath key, in ReadOnlySpan<byte> data, IBatchContext batch);
 
     void Report(IReporter reporter, IPageResolver resolver, int level);
+
+    void Destroy(IBatchContext batch, in NibblePath account);
```
Scooletz, line 33: I'd name it `prefix` as this will be sliced on each level
```suggestion
    void Destroy(IBatchContext batch, in NibblePath prefix);
```

src/Paprika/Store/StorageFanOutPage.cs
```diff
@@ -102,6 +102,24 @@ public void Report(IReporter reporter, IPageResolver resolver, int level)
...
+
+        // Remove the cached
+        batch.IdCache.Remove(prefix.UnsafeAsKeccak);
+        var index = GetIndex(prefix);
+        var addr = Data.Addresses[index];
+        if (addr.IsNull)
+        {
+            // recycleForReuse
+        }
+        TNext.Wrap(batch.GetAt(addr)).Destroy(batch, prefix.SliceFrom(ConsumedNibbles));
```
Scooletz, line 121: This requires walking through the `Map` with `EnumerateAll` and checking whether there's an entry that matches the prefix. On match, `map.Delete(item)`. This should be on each level can cache things. You could even extend `SlottedArray` with a `Delete(in NibblePath prefix)` that would the scan for you?

src/Paprika/Store/FanOutList.cs
```diff
@@ -74,4 +74,19 @@ public void Report(IReporter reporter, IPageResolver resolver, int level)
...
+    public void Destroy(IBatchContext batch, in NibblePath prefix)
+    {
+        // Destroy the Id entry about it
+        Set(prefix, ReadOnlySpan<byte>.Empty, batch);
+
+        // Destroy the account entry
+        // SetAtRoot<FanOutPage>(batch, account, ReadOnlySpan<byte>.Empty, ref Data.StateRoot);
+
+        // Remove the cached
+        batch.IdCache.Remove(prefix.UnsafeAsKeccak);
```
Scooletz, line 86: I think, the removal from `batch.IdCache` should be only in the `RootPage`, once the mapping is removed, we're ok and requires no more removals.

Scooletz (commented):
Some comments.

src/Paprika/Store/StorageFanOutPage.cs
shubham-sonthalia: Added some change. Haven't created a new method for deleting by prefix. Will do it once the logic is correctly implemented. Have a look - @Scooletz

src/Paprika/Store/Page.cs
shubham-sonthalia: Done

src/Paprika/Store/StorageFanOutPage.cs
```diff
@@ -107,18 +107,31 @@ public void Destroy(IBatchContext batch, in NibblePath prefix)
...
-        // Remove the cached
-        batch.IdCache.Remove(prefix.UnsafeAsKeccak);
-        var index = GetIndex(prefix);
-        var addr = Data.Addresses[index];
-        if (addr.IsNull)
+        var map = new SlottedArray(Data.Data);
+        foreach (var item in map.EnumerateAll())
         {
-            // recycleForReuse
+            if (item.Key.Equals(prefix))
```
Scooletz, line 113: Not equals but `StartsWith`. I think this method is missing. If it is, it can be implemented within `NibblePath` using `FindFirstDifferentNibble` and checking for the length

src/Paprika/Store/StorageFanOutPage.cs
```diff
@@ -107,18 +107,31 @@ public void Destroy(IBatchContext batch, in NibblePath prefix)
...
+            {
+                map.Delete(item);
+                var sliced = prefix.SliceFrom(ConsumedNibbles);
```
Scooletz, line 116: This loop over indexes should be separate. Otherwise you'll perform the check for each item.  Conceptually, I'd follow:

1. scan map and delete what is needed
2. propagate down

The first, could be moved to the map as discussed before so that this is a single call. The 2nd, it's specific for a page type.

Scooletz (commented):
Provided some comments. I think, the easiest way to follow would be to have `IPageWithData` require the `DeleteByPrefix` and then follow with all the implementations. It would allow for seamless propagation in all the cases as well.

src/Paprika/Store/StorageFanOutPage.cs
shubham-sonthalia: Is the Equal method defined in NibblePath doing it differently? - (https://github.com/NethermindEth/Paprika/blob/8b91e46d65eb296ba96dc2004a847a9504feb68c/src/Paprika/Data/NibblePath.cs#L4)

src/Paprika/Store/StorageFanOutPage.cs
shubham-sonthalia: Pls check now.

src/Paprika/Store/RootPage.cs
```diff
@@ -224,8 +224,7 @@ public void Destroy(IBatchContext batch, in NibblePath account)
         // Remove the cached
         batch.IdCache.Remove(account.UnsafeAsKeccak);
 
-        // TODO: there' no garbage collection for storage
-        // It should not be hard. Walk down by the mapped path, then remove all the pages underneath.
+        Data.Storage.Destroy(batch, account.SliceFrom(2));
```
Scooletz, line 227: Why slice here? Isn't it `FanOutList` responsibility to slice off?

src/Paprika/Store/StorageFanOutPage.cs
```diff
@@ -102,6 +102,30 @@ public void Report(IReporter reporter, IPageResolver resolver, int level)
             }
         }
     }
+    public void Destroy(IBatchContext batch, in NibblePath prefix)
+    {
+        // Destroy the Id entry about it
+        Set(prefix, ReadOnlySpan<byte>.Empty, batch);
```
Scooletz, line 108: Why call `Set` here? Isn't it handled by the following `DeleteByPrefix`?

src/Paprika/Store/StorageFanOutPage.cs
```diff
@@ -102,6 +102,30 @@ public void Report(IReporter reporter, IPageResolver resolver, int level)
...
+        var map = new SlottedArray(Data.Data);
+        map.DeleteByPrefix(prefix);
+        var sliced = prefix.SliceFrom(ConsumedNibbles);
+        var index = GetIndex(sliced);
+        ref var addr = ref Data.Addresses[index];
+        Page child;
+        if (addr.IsNull)
+        {
+            //recycle for reuse
+            var indexOfPrefix = GetIndex(prefix);
```
Scooletz, line 119: Why do anything while child is null? No need to do it.

src/Paprika/Store/StorageFanOutPage.cs
```diff
@@ -102,6 +102,30 @@ public void Report(IReporter reporter, IPageResolver resolver, int level)
...
         }
     }
+    public void Destroy(IBatchContext batch, in NibblePath prefix)
+    {
+        // Destroy the Id entry about it
+        Set(prefix, ReadOnlySpan<byte>.Empty, batch);
+
+        var map = new SlottedArray(Data.Data);
+        map.DeleteByPrefix(prefix);
+        var sliced = prefix.SliceFrom(ConsumedNibbles);
```
Scooletz, line 112: slice can be done only if there's a child, move donw?

src/Paprika/Store/StorageFanOutPage.cs
Scooletz: Destroy should `COW` like `Set` does. The `batchId` check should be done to not write at readonly pages.

src/Paprika/Store/Page.cs
```diff
@@ -29,6 +29,8 @@ public interface IPageWithData<TPage> : IPage
     Page Set(in NibblePath key, in ReadOnlySpan<byte> data, IBatchContext batch);
 
     void Report(IReporter reporter, IPageResolver resolver, int level);
+
+    void Destroy(IBatchContext batch, in NibblePath prefix);
```
Scooletz, line 33:
```suggestion
    Page Destroy(IBatchContext batch, in NibblePath prefix);
```

Destroy can COW, should return the page.

src/Paprika/Store/StorageFanOutPage.cs
Scooletz: You don't want to equal, you want to destroy also all the merkle nodes that have a prefix.

shubham-sonthalia: Okay, something like this  - 

    public bool StartsWith(in NibblePath other)
    {
        return FindFirstDifferentNibble(other) > 0;
    }

src/Paprika/Store/StorageFanOutPage.cs
shubham-sonthalia: Corrected.

src/Paprika/Store/StorageFanOutPage.cs
shubham-sonthalia: Done.

src/Paprika/Store/StorageFanOutPage.cs
```diff
@@ -102,6 +102,32 @@ public void Report(IReporter reporter, IPageResolver resolver, int level)
...
+    {
+        if (Header.BatchId != batch.BatchId)
+        {
+            // the page is from another batch, meaning, it's readonly. Copy
+            var writable = batch.GetWritableCopy(page);
+            return new StorageFanOutPage<TNext>(writable).Set(prefix, ReadOnlySpan<byte>.Empty, batch);
+        }
+
+        // Destroy the Id entry about it
+        // Set(prefix, ReadOnlySpan<byte>.Empty, batch);
```
shubham-sonthalia, line 115: @Scooletz  I have commented out the Set call here. Wanted to confirm if I need to add it to the DeleteByPrefix method?

src/Paprika/Store/StorageFanOutPage.cs
```diff
@@ -102,6 +102,32 @@ public void Report(IReporter reporter, IPageResolver resolver, int level)
             }
         }
     }
+    public Page Destroy(IBatchContext batch, in NibblePath prefix)
+    {
+        if (Header.BatchId != batch.BatchId)
+        {
+            // the page is from another batch, meaning, it's readonly. Copy
+            var writable = batch.GetWritableCopy(page);
+            return new StorageFanOutPage<TNext>(writable).Set(prefix, ReadOnlySpan<byte>.Empty, batch);
```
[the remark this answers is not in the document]
Scooletz: Each method that modified things, needs to COW. To Copy-On-Write, you do the writable thing and then you call the same method again, so that it should be `Destroy` called here.

src/Paprika/Store/StorageFanOutPage.cs
Scooletz: On each level, if a page has a map, the map should have all the prefixes removed.

Scooletz (commented):
Added some comments. Why no tests? Having tests would be beneficial from the start. It would be much easier to discuss scenarios, over test suite and then follow with the implementation.

src/Paprika/Data/NibblePath.cs
```diff
@@ -517,7 +517,10 @@ public bool Equals(in NibblePath other)
 
         return FindFirstDifferentNibble(other) == Length;
     }
-
+    public bool StartsWith(in NibblePath other)
+    {
+        return FindFirstDifferentNibble(other) > 0;
```
Scooletz, line 522: If two path of length `3` are different at `2`, they do not start with each other, do they? 👀

shubham-sonthalia:
I will add test cases. I am travelling this week with limited access to network. Is it okay if I take some time for this?

Scooletz:
It's more than ok @shubham-sonthalia ❤️ Pinge me once you get back to this and I'll keep you here in the loop if needed.

src/Paprika/Data/NibblePath.cs
shubham-sonthalia: I have changed the implementation to  - 

return FindFirstDifferentNibble(other) == Length 

Does this look good?

src/Paprika/Store/StorageFanOutPage.cs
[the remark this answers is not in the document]
shubham-sonthalia: Have implemented it like this - 

`if (Header.BatchId != batch.BatchId)
        {
            // the page is from another batch, meaning, it's readonly. Copy
            var writable = batch.GetWritableCopy(page);
            return new StorageFanOutPage<TNext>(writable).Destroy(batch, prefix);
        }
`
Looks good?

shubham-sonthalia:
Hi @Scooletz , I am available for further discussion on this. I am working on the test cases for this. Can you check the other two issues that I resolved?

src/Paprika/Store/StorageFanOutPage.cs
shubham-sonthalia: Hey, does this point still need to be addressed?

shubham-sonthalia:
> @shubham-sonthalia The changes look ok. Tests, we need them. Also, some conflicts. How can I help?

Hi @Scooletz For Tests, I needed some help. I need some reference in code to put proper Assert conditions. Can you nudge me towards the right approach?

shubham-sonthalia:
HI @Scooletz , have created this new PR for this issue to avoid merge conflicts and have a fresh discussion on the changes - https://github.com/NethermindEth/Paprika/pull/282. 

i have added test cases for two methods. Do let me know if any concerns.

Scooletz:
Closing this in favor of #282 then. Will review the other today. Thank you 🙇

PR #264, opened 2024-02-29
Leaf + overflow

Scooletz:
This PR introduces a new page called `LeafPage` that is capable to offload bigger chunks of data to one of its `LeafOverflowPage`. The idea is to separate the overflow behavior from the `DataPage` and use a separate set of `LeafPage` + `OverflowPage` for this purpose. Additionally, the `OverflowPage` provides with almost the whole page (- header which is 8 bytes) for the slotted array which should provide some additional space.

To use the `LeafPage`s fully, even without overflowing behavior, before flushing the biggest nibble, from now on `DataPage` will test all the leaf pages that it has, whether they can accept a write. This should create slightly unbalanced trie with pages that are much more full

Additionally, this PR introduces:

1. `TreeView` that is helpful to see the database layout in tests.

### Results

| Network  | RocksDb   | Paprika (before)  | Paprika (after) | Size reduction |
|---|---|---|---|---|
|  Holesky  | 6.7GB  | 7.3GB  | tbd  | tbd  |
| Mainnet  |  184G | 165GB | 159GB | tbd  |


Scooletz:
Maybe offload to the page only once the original is filled with data? If it's not, why even bother offloading big items?

PR #267, opened 2024-03-04
Dense slotted array

Scooletz:
This PR changes the way `SlottedArray` generates the hash. It makes it in a way that nibbles used for hash creation are removed from the key so that key is much smaller. This should make the db much more compact and the trees less deep.

An additional change that was introduced after running benchmarks is using both prefix and suffix for longer keys. As tree in Paprika has a write-caching behavior, heavily written storage slots for same accounts could dominate the map with their prefixes. In that case, the search would be terribly slow (shown by benchmarks). To address it, when a key is longer than 4 nibbles, 2 first nibbles and 2 last nibbles are extracted as hash. This should limit the simple collisions that can be caused by data not being flushed deep into the tree.

### Benchmarks

Benchmarks showing that using prefix/suffix nibbles can help a lot where the data share the prefix (described above).

| Method                              | Mean     | Error    | StdDev   | Median   |
|------------------------------------ |---------:|---------:|---------:|---------:|
| Write_whole_page_of_data            | 21.16 us | 0.420 us | 1.178 us | 21.34 us |
| Read_existing_keys_prefix_different | 21.76 us | 0.200 us | 0.187 us | 21.78 us |
| (before) Read_existing_keys_suffix_different | **903.65 us** | 20.705 us | 61.048 us | 931.70 us |
| (after) Read_existing_keys_suffix_different | **19.29 us** | 0.520 us | 1.533 us | 19.48 us |
| Read_nonexistent_keys               | 21.01 us | 1.031 us | 2.992 us | 22.83 us |

### Results

The results that are shown are for different versions of this approach

| Network  | RocksDb   | Paprika (before)  | Paprika (after) | Size reduction | Version |
|---|---|---|---|---|---|
|  Holesky  | 6.7GB  | 7.3GB  | 7GB  | ~4%  | at 3d5aee6, 3 nibbles extracted
| Mainnet  |  184G | 165GB | 159GB | ~4%  | at 3d5aee6, 3 nibbles extracted
| Mainnet  |  184G | 165GB | 159GB | ~4%  | with 4 nibbles extracted

#### Mainnet

image

### DIFF
--- a/src/Paprika.Benchmarks/Paprika.Benchmarks.csproj
+++ b/src/Paprika.Benchmarks/Paprika.Benchmarks.csproj
@@ -9,8 +9,8 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <PackageReference Include="BenchmarkDotNet" Version="0.13.7" />
-      <PackageReference Include="BenchmarkDotNet.Diagnostics.dotTrace" Version="0.13.7" />
+      <PackageReference Include="BenchmarkDotNet" Version="0.13.12" />
+      <PackageReference Include="BenchmarkDotNet.Diagnostics.dotTrace" Version="0.13.12" />
     </ItemGroup>
 
     <ItemGroup>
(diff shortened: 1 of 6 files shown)


Scooletz:
image

A snippet from syncing. We'll see.

benaadams (approved):

PR #269, opened 2024-03-05
Trie for Merkle

Scooletz:
A highly speculative PR that reintroduced `TrieNode` in an unmanaged, alloc friendly way. The aim is to amortize the constant decoding/encoding of values when marking path as dirty. A similar semantics to what is done with `TrieNode` could be used, meaning, dirty, unresolved etc.


Scooletz:
Doesn't look like an issue atm. May revisit later.

PR #271, opened 2024-03-05
Delayed metrics introduced

Scooletz:
This PR introduces a delayed metrics reporting for xor misses. If a reporter is registered that does a lot of work on the `Record(...)` like Prometheus for .NET (see: [OnMeasurementRecorded](https://github.com/prometheus-net/prometheus-net/blob/60e9106a83ff1274fec0022c37366f04822b1d1b/Prometheus/MeterAdapter.cs#L126-L215)), then it can impact the the cases where xor reports positively quite a lot. After introduction of the `Delayed`, the value is reported once at the end of the lifetime of the block state.

Below, two snippets from the profiling

Storage computation
State computation

### DIFF
--- a/src/Paprika/Chain/Blockchain.cs
+++ b/src/Paprika/Chain/Blockchain.cs
@@ -429,7 +429,7 @@ public class Blockchain : IAsyncDisposable
         /// </summary>
         private PooledSpanDictionary _preCommit = null!;
 
-        private readonly DelayedMetrics.ICounter<long> _xorMissed;
+        private readonly DelayedMetrics.DelayedCounter<long, DelayedMetrics.LongIncrement> _xorMissed;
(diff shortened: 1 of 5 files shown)


src/Paprika/Chain/Blockchain.cs
```diff
@@ -429,6 +429,7 @@ private class BlockState : RefCountingDisposable, IWorldState, ICommit, IProvide
         /// </summary>
         private PooledSpanDictionary _preCommit = null!;
 
+        private readonly DelayedMetrics.ICounter<long> _xorMissed;
```
benaadams, line 432: Does it have to be interface (so indirect call per call)? Could it be a concrete type?

Scooletz: Fixed @benaadams

benaadams (approved):

PR #272, opened 2024-03-05
ScatterGather balancing

benaadams:
Increase the number of batches as they look to periodically become very imbalanced

image

PR #273, opened 2024-03-05
Merkle scatter gather simplified

Scooletz:
This PR touches the same area as #272 delegating now fully the work to the underlying partitioner of PLINQ. It may cause some spikes as items are potentially not distributed ideally, but should make run simpler and less skewed in cases where the distributions is similar.


benaadams (approved):

benaadams:
Test timeout https://github.com/NethermindEth/Paprika/pull/274

benaadams (approved):

PR #274, opened 2024-03-05
Increase test timeout

benaadams:
Increase test timeout


Scooletz (approved):
As long as it's bounded 🔥

PR #275, opened 2024-03-06
Prefetching Merkle

Scooletz:
This is an initial sketch of a prefetcher, capable of prefetching the Merkle structure behind the scenes. This is done so that once `Commit` is called, all is set and is ready to go.

Doesn't work at the moment.


Scooletz:
Replaced by #279

PR #278, opened 2024-03-12
Optimize BitVector

benaadams:
- Use machine word sized referencing (long)
- Compare if 0 rather than selector
- Use shift wrapping rather than additional mask
- Make getters as `readonly`


Scooletz (commented):
🌶️

PR #279, opened 2024-03-13
Merkle Prefetcher

Scooletz:
This is an initial sketch of a prefetcher, capable of prefetching the Merkle structure behind the scenes. This is done so that once Commit is called, all is set and is ready to go.


Scooletz:
Will be replaced by another one.

PR #286, opened 2024-03-21
Merkle branch rlp memoization

Scooletz:
This PR changes the memoization scheme, from memoizing Keccak value of a branch node within it to branches memoizing Keccaks of its children. This makes it much more similar to [`TrieNode` of Nethermind](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Trie/TrieNode.cs). The reason for this is to reduce the read amplification occurring, when a branch node is read and it has none of its children cached. This change should cut the number of reads greatly.

The expected decrease of reads is at level of 16, as majority of the recalculated branches are full and have 16 children. Please note, that for bottom branches, it was leafs that were read as well as the underlying values. This is no longer the case with this memoization.

### Db size overhead

There's a caveat, as with this memoization, there's an expected db size increase that might happen. As all the children will have their Keccaks memoized, there is more data to be stored. To make it work compression is applied just before hitting the db, removing all the entries that belong to the non-existing children. The hydration happens on the reads. 

This means that, for example, a branch with two leafs, will be stored on 1 + 2 + 32*2 = 67 bytes. 

- 1 - the node header
- 2 - the branch bits
- 32 * 2 - two Keccaks memoized

#### Branch optimization

An additional optimization is introduced, so that if a branch has only two children and is on even level of the trie (either storage or state), it won't have its children's Keccak memoized. The reason for that is that if this branch is marked as dirty, it will have once of its children invalidated on update and one removed on delete. The addition will move it to a branch with 3. This means that the read amplification for such a case is not that big while the savings on the disk can be. 

The reason, why every other level is forgotten, is to prevent an attack when a malicious actor creates a log tree with branches that have only 2 children each. With forgetting every other level, the reads won't go out of control.


Scooletz:
https://github.com/NethermindEth/Paprika/pull/286#issuecomment-2012952956, meaning non-encoding zeros, provided at [44a57cf](https://github.com/NethermindEth/Paprika/pull/286/commits/44a57cf008fbd98ac34591e88b1d722acd4d0ae5)

Scooletz:
The final output 

image

Scooletz:
As #293 has been merged to main, the expected size of the imported database is much smaller.

PR #287, opened 2024-03-21
Optimize structures

benaadams:
- Removed 2 `Span<T>` fields from `SlottedArray`
- Use LayoutKind.Sequential rather than LayoutKind.Explicit when possible as `Explicit` is a deoptimization for Jit (due to overlapping risk)
- Don't throw directly inside hot paths; but throw from local function so `throw` code is not cold code inside asm
- More `SkipLocalsInit` where stackalloc is used


src/Paprika/Data/SlottedArray.cs
```diff
@@ -450,7 +469,7 @@ private Span<byte> GetSlotPayload(ref Slot slot)
     /// </summary>
     public static ushort HashForTests(in NibblePath key) => Slot.PrepareKey(key, out _, out _);
 
-    [StructLayout(LayoutKind.Explicit, Size = Size)]
+    [StructLayout(LayoutKind.Sequential, Pack = sizeof(byte), Size = Size)]
     private struct Slot
```
Scooletz, line 473: This is persistent data structure. I believe we should go with `Explicit` here. We might come up with an attribute for such cases or a comment.

src/Paprika/Data/SlottedArray.cs
```diff
@@ -24,15 +25,32 @@ namespace Paprika.Data;
...
     public SlottedArray(Span<byte> buffer)
     {
-        _raw = buffer;
-        _header = ref Unsafe.As<byte, Header>(ref _raw[0]);
+        _header = ref Unsafe.As<byte, Header>(ref MemoryMarshal.GetReference(buffer));
         _data = buffer.Slice(Header.Size);
-        _slots = MemoryMarshal.Cast<byte, Slot>(_data);
+    }
+
+    private readonly ref Slot this[int index]
```
Scooletz, line 35: This is cool.

src/Paprika/Store/AbandonedPage.cs
```diff
@@ -24,7 +24,7 @@ namespace Paprika.Store;
 
     private unsafe ref Payload Data => ref Unsafe.AsRef<Payload>(page.Payload);
 
-    [StructLayout(LayoutKind.Explicit, Size = Size)]
+    [StructLayout(LayoutKind.Sequential, Pack = sizeof(byte), Size = Size)]
```
Scooletz, line 27: This is also persistent thing. Cannot go sequential I believe.

src/Paprika/Store/Page.cs
```diff
@@ -54,7 +54,7 @@ public static class PageExtensions
 /// <summary>
 /// The header shared across all the pages.
 /// </summary>
-[StructLayout(LayoutKind.Explicit, Pack = 1, Size = Size)]
+[StructLayout(LayoutKind.Sequential, Pack = sizeof(byte), Size = Size)]
```
Scooletz, line 57: Persistent

src/Paprika/Store/LeafOverflowPage.cs
```diff
@@ -15,15 +15,16 @@ namespace Paprika.Store;
 
     private ref Payload Data => ref Unsafe.AsRef<Payload>(page.Payload);
 
-    [StructLayout(LayoutKind.Explicit, Size = Size)]
+    [InlineArray(Size)]
+    [StructLayout(LayoutKind.Sequential, Pack = sizeof(byte), Size = Size)]
```
Scooletz, line 19: Persistent

src/Paprika/Store/RootPage.cs
```diff
@@ -240,14 +240,14 @@ public void Destroy(IBatchContext batch, in NibblePath account)
     }
 }
 
-[StructLayout(LayoutKind.Explicit, Size = Size, Pack = 1)]
+[StructLayout(LayoutKind.Sequential, Pack = sizeof(byte), Size = Size)]
```
Scooletz, line 243: Persistent

Scooletz (commented):
Looks promising. The `this[index]` for `SlottedArray` is 😋 .
One thing that is repeated is the `Sequential` for structs that are persistent. We should go with `Explicit` I believe as otherwise a simple move of a field or a runtime change layout will kill the db.

src/Paprika/Data/SlottedArray.cs
benaadams: Sequential is fine, just don't move fields 🙂

src/Paprika/Store/AbandonedPage.cs
benaadams: Sequential is fine, just don't move fields 🙂

src/Paprika/Store/LeafOverflowPage.cs
benaadams: Sequential is fine, just don't move fields 🙂

src/Paprika/Store/RootPage.cs
benaadams: Sequential is fine, just don't move fields 🙂

Scooletz (approved):
The versioning of the structure is parially done with #288. Let's go.

PR #289, opened 2024-03-22
Optimize DataPage lookup

benaadams:
- Change DataPage.TryGet into single method iteration rather than multi-call recursion, which means the stack is only prepped once rather than 7 times if the data pages are 7 deep
- Pass params <= pointer sized first (better for abi, more register vs stack use)
- Split long methods with distinct early exits into multiple; as all the stack prep is done on method entry whether or not its needed. So splitting into multiple means methods only do as much stack prep as they need for that chunk of code rather than always doing it for everything even if only the first bit is called.
- Unsafe.As cast Pages in-place rather than going via constructor and using more stack

### DIFF
--- a/src/Paprika.Tests/Store/DataPageTests.cs
+++ b/src/Paprika.Tests/Store/DataPageTests.cs
@@ -46,7 +46,7 @@ public class DataPageTests : BasePageTests
             Keccak search = default;
             BinaryPrimitives.WriteInt32LittleEndian(search.BytesAsSpan, j);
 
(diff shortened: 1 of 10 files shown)


Scooletz (approved):
Really cool 😍 

The looping over is awesome. 

You added `in` for the `NibblePath` in `TryGet` which I failed to do previously as there was some case where it was not possible to have it (can't remember now). This makes it also a pointer size so as far as I understand, the moving of the batch does not change the actual where pointer sized should go first? 🤔 

Anyway, looks awesome and I'm merging it in.

PR #290, opened 2024-03-23
Constructing PageAccessor and using it to create _fingerprints

### LINKED ISSUE #192: Clean allocations in Xor
#191 introduced great performance gains but added a bit of allocs. It can be removed and pool can be leveraged more. The list of allocations:

- [x] `fp` - in #259 
- [ ] `_fingerprints` (see: [code](https://github.com/NethermindEth/Paprika/blob/cda38d382fd823b4e2dacde6573db344f61d2af5/src/Paprika/Utils/Xor8.cs#L28)) and replacing them with something based on the `BufferPool`

The last one requires constructing a page accessor that would allow to use multiple pages/buffers from the pool and dispose them properly at the end. It could something as simple as `Page[]` as it would already provide a good fanout 1 item -> 4kb of memory.

shubham-sonthalia:
Resolves #192

### DIFF
--- a/src/Paprika/Utils/Xor8.cs
+++ b/src/Paprika/Utils/Xor8.cs
@@ -6,6 +6,7 @@ using System.Numerics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.InteropServices.ComTypes;
+using Paprika.Store;
 
 namespace Paprika.Utils;
 
@@ -29,6 +30,8 @@ public class Xor8
     private readonly ulong _seed;
     private readonly byte[] _fingerprints;
 
+    private PageAccessor pageAccessor;
+
     private static int GetArrayLength(int size) => (int)(Offset + (long)FactorTimes100 * size / 100);
 
     public Xor8(HashSet<ulong> keys)
@@ -46,6 +49,7 @@ public class Xor8
 
         int reverseOrderPos;
         ulong seed;
+
 
     MainLoop:
         do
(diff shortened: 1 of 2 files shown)


Scooletz (changes requested):
This solution provides a managed solution that we want to go aware from. There's the `BufferPool` already that provides `4kb` chunks that are used by `PooledSpanDictionary` and others. Would it be possible to have it migrated to use this option? Otherwise, these chunks cannot be shared and we land in a world where different parts of Paprika used different pools.

src/Paprika/Utils/Xor8.cs
```diff
@@ -216,6 +221,10 @@ private int GetHash(ulong key, ulong seed, int index)
 
     private static int Fingerprint(ulong hash) => (int)(hash & ((1 << BitsPerFingerprint) - 1));
 
+    ~Xor8()
```
benaadams, line 224: Finalizers are bad for performance, would be better to use IDisposable or similar

shubham-sonthalia:
> This solution provides a managed solution that we want to go aware from. There's the `BufferPool` already that provides `4kb` chunks that are used by `PooledSpanDictionary` and others. Would it be possible to have it migrated to use this option? Otherwise, these chunks cannot be shared and we land in a world where different parts of Paprika used different pools.

Sure, will work on integrating the BufferPool instead of the new class.

src/Paprika/Utils/Xor8.cs
shubham-sonthalia: Yes, will remove it.

Scooletz:
#345 replaces this providing an ORable filter that uses the unmanaged memory

PR #292, opened 2024-03-27
Much better statistics

Scooletz:
This PR improves the statistics reporting for the database, providing better split in data and less ceremony in regards to the gathering page usage across different page types. Additionally, it separates the level the page is at from the number of nibbles that are consumed by the given structure. This allows to efficiently find what is a Merkle (a not-full path) and what is not (64 nibbles in path) and report it properly.


Scooletz:
Too hard to port back. Will be a part of #286

PR #296, opened 2024-03-29
Batching file writes

Scooletz:
Paprika uses memory mapped files for its storage. To persist changes, it issues writes without `MSYNC` or `FlushViewOfFile` which have terrible performance, but with a regular `RandomAccess.WriteAsync`. As the underlying file is mapped, issuing writes will write through the cache and keep coherency of the view. This PR alters the way writes are scheduled to disk and leverages a potential consecutive pages. Once a longer slice is detected, up to 64 pages will be batched together.

Measurements show a great reduction of `ScheduleWrites` and moving the weight to where it belongs, the FSYNC call.

This usually should not impact the speed, but can lead to blocks accumulating in the flusher queue. By speeding up the `FlusherTask` the right of high saturation is getting lower.

image

image

Clearly, there's some potential to improve page collocation when reused, but this is a step in a good direction to limit the number of syscall and make it more visible.

### DIFF
--- a/src/Paprika.Runner/Program.cs
+++ b/src/Paprika.Runner/Program.cs
@@ -51,7 +51,7 @@ public static class Program
         new(50_000, 1000, 11 * Gb, true, TimeSpan.FromSeconds(5), false, false);
 
     private static readonly Case DiskSmallFlushFile =
-        new(50_000, 1000, 32 * Gb, true, TimeSpan.FromSeconds(5), true, false);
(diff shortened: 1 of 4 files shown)


Scooletz:
Looks like the WriteFile reduction should be bigger, to investigate.

Scooletz:
It looks the abandoned page mechanism is responsible for this scattered writes. Even if this is the case, this PR makes it a bit better, by issuing ~30% less writes.

PR #297, opened 2024-03-30
Xor8: Drop 1 arraypool rent+return, array copy

benaadams:
Is essentially the final array, so why rent for it then allocate and copy to it. Why not just start with the alloc

### DIFF
--- a/src/Paprika/Utils/Xor8.cs
+++ b/src/Paprika/Utils/Xor8.cs
@@ -153,7 +153,7 @@ public class Xor8
         t2 = null;
 
         _seed = seed;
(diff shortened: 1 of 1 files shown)


Scooletz:
I was not sure when porting whether this is the case @benaadams . Will take on it on Tuesday.

benaadams:
> I was not sure when porting whether this is the case @benaadams . Will take on it on Tuesday.

allocate a1
use a1
allocate a2 
copy a1 to a2
decallocate a1

skip the copy and second alloc it becomes

allocate a1
use a1

Scooletz (approved):

PR #301, opened 2024-04-04
Further optimize SlottedArray

benaadams:
Doesn't need to be two object references + length; can be 1 object ref + data length


Scooletz:
@benaadams Looks like not that much gains and the `Hash_collisions` benchmark went through the roof? Would you have some time to take a look at it?

Scooletz (changes requested):
@benaadams Looks like not that much gains and the `Hash_collisions` benchmark went through the roof? Would you have some time to take a look at it? I've added results to the description

PR #306, opened 2024-04-08
Page 8kb

Scooletz:
This PR modifies `SlottedArray` so that it can use 13 bits instead of 12 to store addresses within page. This allows to move from 4kb to 8kb pages. Having bigger, 8kb pages, can impact the storage in the following manner. Can as it's yet to be measured:

1. 8kb pages are 2 times bigger than OS pages and as `SlottedArray` stores hashes in its `Slots`, it's possible to search through just prefixes without touching the data if the data are not there
1. 8kb pages allow to amortize more and gather more data before flushing down
1. root page management must be changed:
 1. root can store more as its 8kb now
 2. root is not atomically storeable. Should be changed so that it's written as usual and additionally, at `DbAddress.Null` (offset 0), the actual last root number is stored

## Page Size considerations

Memory page sizes vary significantly across different operating systems and CPU architectures, including ARM, x64, and Apple's proprietary chips like the M3. Below is a breakdown by the mentioned categories:

### ARM Architecture
- **Generic ARM Systems**: Typically use a default page size of 4 KB (kilobytes), with support for larger page sizes like 64 KB, depending on system configuration.
- **Apple Silicon (Including M1, M2, and presumable M3)**: Inherits ARM's flexible page size support with a default page size of 16 KB for user space and 4 KB for kernel space. Official documentation for M3 specifics might need to be consulted for confirmation.

### x64 Architecture (AMD64/Intel 64)
- **General x64 Systems**: The typical page size is 4 KB, with support for large pages that are 2 MB or even 1 GB, depending on the system's configuration and operating system support.
- **Linux on x64**: Supports 4 KB default page sizes with the possibility to use huge pages of 2 MB or 1 GB.
- **Windows on x64**: Uses a standard 4 KB page size, with options for large pages at 2 MB.

### Operating Systems
- **Linux**: Supports 4 KB default page sizes with the option for huge pages (2 MB or 1 GB).
- **Windows**: Uses a default page size of 4 KB, with support for large pages of 2 MB.
- **macOS (Running on Intel and Apple Silicon)**: Supports a default page size of 4 KB on Intel-based systems and 16 KB on Apple Silicon for user applications, with 4 KB pages in kernel mode.

### Notes
While the default page sizes are commonly 4 KB across many systems, modern operating systems and CPUs support a variety of page sizes to optimize for different use cases, such as reducing page table sizes or minimizing TLB (Translation Lookaside Buffer) misses. The exact page size used can depend on various factors, including the operating system's memory management policies, the CPU's capabilities, and the application's needs. For the most accurate and up-to-date information, consulting the specific CPU architecture or operating system documentation is recommended.


benaadams (approved):
Tests need changing I think? As expecting smaller sizes/different data format

Scooletz:
> Aside Apple M3 CPU's page size is 16KB and cacheline size is 128B

What I can mention is that `LMDB` uses `4kb` while `Raven` does `8kb`. Both use mmap. I'm considering making `RootPage` much much bigger as a separate thing, but this would require a change in flushing.

> Tests need changing I think? As expecting smaller sizes/different data format

Yes. This requires more testing in general, like running a full import to see the behavior. My gut feeling, with the data being now of different sizes totally and branches being 512bytes long, it will make the tree more adaptive and smaller. But more tests needed.

Scooletz:
Closing as `SlottedArray` allows now 8kb pages but it does not provide much benefit. This allowance may be used later though. I any case, this PR is not needed as it has been implemented.

PR #307, opened 2024-04-08
Robustify RefCountingDisposable

benaadams:
Currently if 2 threads simultaniously try to Aquire a disposed object one may suceed

1. Starting value = -1
2. Thread 1: increment = 0
3. Thread 2: increment = 1
4. Thread 1, fail: decrement = 0
5. Thread 2, suceeded => invalid state

### DIFF
--- a/src/Paprika/Utils/RefCountingDisposable.cs
+++ b/src/Paprika/Utils/RefCountingDisposable.cs
@@ -128,7 +128,7 @@ public abstract class RefCountingDisposable : IDisposable
         return leases == Disposing ? "Disposed" : $"Leases: {leases}";
     }
 
-    [StructLayout(LayoutKind.Explicit, Size = 64 + sizeof(long))]
+    [StructLayout(LayoutKind.Explicit, Size = 128)]
     private struct PaddedValue
     {
         [FieldOffset(64)]
--- a/src/Paprika/Utils/RefCountingDisposable.cs
+++ b/src/Paprika/Utils/RefCountingDisposable.cs
@@ -9,13 +9,13 @@ namespace Paprika.Utils;
 /// </summary>
 public abstract class RefCountingDisposable : IDisposable
 {
-    private const int Initial = 1;
+    private const int Single = 1;
     private const int NoAccessors = 0;
     private const int Disposing = -1;
 
     private PaddedValue _leases;
 
-    protected RefCountingDisposable(int initialCount = Initial)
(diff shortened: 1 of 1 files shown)


src/Paprika/Utils/RefCountingDisposable.cs
```diff
@@ -11,13 +12,12 @@ public abstract class RefCountingDisposable : IDisposable
     private const int Initial = 1;
     private const int NoAccessors = 0;
     private const int Disposing = -1;
-    private const int DisposingBarrier = 0;
 
-    private int _leases;
+    private PaddedValue _leases;
```
Scooletz, line 16: I see the padding here. I know what it does. The thing is that this is the abstract class and what I'm thinking is whether is needed. WDYT?

src/Paprika/Utils/RefCountingDisposable.cs
```diff
@@ -57,23 +73,65 @@ protected bool TryAcquireLease()
...
-            if (Interlocked.CompareExchange(ref _leases, Disposing, NoAccessors) == NoAccessors)
+            var prev = Interlocked.CompareExchange(ref _leases.Value, current - 1, current);
+            if (prev != current)
+            {
+                current = prev;
+                // Add PAUSE instruction to reduce shared core contention
+                Thread.SpinWait(1);
+                continue;
+            }
+            if (prev == 1)
```
Scooletz, line 94: Magic const. This could be 
```suggestion
            if (prev == Initial)
```

Scooletz (commented):
Good to go after clarifications @benaadams

src/Paprika/Utils/RefCountingDisposable.cs
benaadams: Changed

src/Paprika/Utils/RefCountingDisposable.cs
benaadams: If it is hevily used changing its value causes anything on its cache line to be invalidated for other CPUs and forces them to be reloaded.

So for example the pointer of `HashSet<ulong> _bloom` would keep getting invaliated in `BlockState`; or pointer of `PooledSpanDictionary _dict` in `ChildCommit`

benaadams: Depends if worried about the memory usage by the extra padding or not

Scooletz (approved):

PR #311, opened 2024-04-10
Faster UnPrepareKey

benaadams:
`UnPrepareKey` used in the enumeration over the map was slow when running `Importer`. This can also impact flushing data down while running the background `FlusherTask` of `Blockchain`. Should speed up data application as well.

#### Before

| Method       | Mean     | Error     | StdDev    | Code Size |
|------------- |---------:|----------:|----------:|----------:|
| EnumerateAll | 7.877 us | 0.1552 us | 0.1725 us |   2,487 B |

#### After

| Method       | Mean     | Error     | StdDev    | Code Size |
|------------- |---------:|----------:|----------:|----------:|
| EnumerateAll | 6.646 us | 0.0893 us | 0.0835 us |   1,756 B |


src/Paprika/Data/NibblePath.cs
```diff
@@ -47,6 +67,12 @@ public static NibblePath Single(byte nibble, int odd)
         return new NibblePath(ref Unsafe.Add(ref singles, nibble), (byte)odd, 1);
     }
 
+    public static NibblePath Dual(byte @byte)
```
Scooletz, line 70: 😅 Next one is triples and quadruples 😜

src/Paprika/Data/NibblePath.cs
```diff
@@ -105,19 +136,16 @@ public static NibblePath FromRawNibbles(ReadOnlySpan<byte> nibbles, Span<byte> w
...
+    public void UnsafeMakeOdd()
     {
-        if (odd == 0)
-            return this;
-
         Debug.Assert(_odd == 0, "Should not be applied to odd");
 
         for (int i = Length; i > 0; i--)
         {
             UnsafeSetAt(i, 0, GetAt(i - 1));
```
Scooletz, line 145: I was thinking that this is the hot spot, especially for lengthy ones.

src/Paprika/Data/SlottedArray.cs
```diff
@@ -177,25 +175,25 @@ public bool MoveNext()
...
             }
 
             return false;
         }
 
-        public Item Current => _current;
+        public readonly Item Current => _current;
 
-        private Item Build()
+        private void Build(out Item value)
```
Scooletz, line 187: Would be great if you could add an enumerator benchmark to `SlottedArrayBenchmarks` to cover this and other changes.

src/Paprika/Data/SlottedArray.cs
```diff
@@ -177,25 +175,25 @@ public bool MoveNext()
...
         {
             ref var slot = ref _map[_index];
             var span = _map.GetSlotPayload(ref slot);
-            var key = Slot.UnPrepareKey(span, slot.Hash, slot.KeyPreamble, _bytes.Span, out var data);
+            var key = Slot.UnPrepareKey(slot.Hash, slot.KeyPreamble, span, _bytes.Span, out var data);
```
Scooletz, line 191: What is the reason for the reorg? Should not big param go first?

src/Paprika/Data/SlottedArray.cs
```diff
@@ -601,50 +599,48 @@ public static ushort PrepareKey(in NibblePath key, out byte preamble, out Nibble
                             (key.GetAt(length - 2) << shift) | key.GetAt(length - 1));
         }
 
-        public static NibblePath UnPrepareKey(ReadOnlySpan<byte> input, ushort hash, byte preamble, Span<byte> workingSet, out ReadOnlySpan<byte> data)
+        [SkipLocalsInit]
+        public static NibblePath UnPrepareKey(ushort hash, byte preamble, ReadOnlySpan<byte> input, Span<byte> workingSet, out ReadOnlySpan<byte> data)
         {
-            var odd = preamble & KeyPreambleOddBit;
             var count = preamble >> KeyPreambleLengthShift;
+            if (count == 0)
```
Scooletz, line 606: The author of #306 will 😭 over this changes. Merge conflict ahead.

Scooletz (commented):
Initial comments. Looks 😋

src/Paprika/Data/SlottedArray.cs
benaadams: https://learn.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-170#calling-convention-defaults

Only 4 parameters are passed by register; for instance that's reduced by 1, for generic reduced by 1 more; anything larger than 8 bytes goes via stack any thing beyond 4 params goes via stack.

image


image

So you want smaller parameters first (including `in` and `ref`) since any large parameters will need to go on stack anyway, so why waste a register position?

Scooletz: This is clearly not how remembered it. Thank you.

Scooletz:
@benaadams I find no performance difference in the iterating scenario.

src/Paprika/Data/NibblePath.cs
benaadams: Updated

PR #315, opened 2024-04-15
Different flush strategy

Scooletz:
This PR introduces different stats counting, tests plus different flushing behavior.


Scooletz:
Won't be pursued.

PR #316, opened 2024-04-16
CPU prefetch

benaadams:
As the CPU does not auto-prefetch across page boundaries, 
Prefetch child page in case we go there next to reduce CPU stalls

### DIFF
--- a/src/Paprika.Tests/Store/BasePageTests.cs
+++ b/src/Paprika.Tests/Store/BasePageTests.cs
@@ -27,6 +27,9 @@ public abstract class BasePageTests
 
         public override Page GetAt(DbAddress address) => _address2Page[address];
 
+        public override void Prefetch(DbAddress address)
+        { }
+
         public override DbAddress GetAddress(Page page) => _page2Address[page.Raw];
 
         public override Page GetNewPage(out DbAddress addr, bool clear)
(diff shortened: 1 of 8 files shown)


Scooletz (commented):
The question that keeps me hanging is why other folks would `MADVISE` then if they can go `PREFETCHh`? Will it additionally hint that it's needed and load through memory because the underlying is mmaped thing? 

Have you checked it with bigger than memory data size that it actually helps?

I do understand the PREFETCH, I do understand MADVISE, I do understand mmap. I miss the intersection of these. Can you elaborate?

benaadams:
> I do understand the PREFETCH, I do understand MADVISE, I do understand mmap. I miss the intersection of these. Can you elaborate?

`MADVISE` (and `PrefetchVirtualMemory` on Windows) will actually cause the page fault in background if one is needed (as that's OS level)

`PREFETCH` will bring memory that's in a mapped page to processor cache and fixup the CPU's TLB (translation lookaside buffer) for the memory translation (which can be expensive)

So will still get page faults if one is going to happen; but the CPU won't get stalled if the memory is there because it doesn't auto-prefetch across page boundaries (it will feed cache with predictive prefetching but stop whenever its going to cross a page)

Scooletz:
I do understand it in the way you described. Why then not `PREFETCH` always as you can assume that the first few bytes (according to spec of CPU that you run on) will be used. Why `MADVISE` at all in this case?

benaadams:
> I do understand it in the way you described. Why then not `PREFETCH` always as you can assume that the first few bytes (according to spec of CPU that you run on) will be used. Why `MADVISE` at all in this case?

`MADVISE` and `PrefetchVirtualMemory` will cause the page fault in background, so can be helpful to page in memory before needed. However is also a syscall and not sure what the hit from post-spectre trampolines is. Sometthing to look at in future.

PR #318, opened 2024-04-18
Storage trees

Scooletz:
This PR changes how `Storage` is handled in Paprika. Previously, the mapping for accounts that have storage assigned natural numbers to them, and used the natural number as the prefix of a storage `NibblePath`. Now, it provides an address of the `StorageRootPage` that can host a limited number of accounts. This changes the dynamic of the storage:

1. For contracts with a big storage trie, the data access can now directly jump to the tip of the Storage trie. It should reduce the number of levels to walk through by ~8 (4 bytes of an integer translates to 8 nibbles that were used as the prefix). 
1. For contracts with a small storage, should be able to set/get data directly from the pointed page slot.
1. For all contracts, their storages should have better locality which should also positively impact the flushing down behavior.

Additional changes:

1. a special encoding for `DbAddress` that for small ranges (up to 0x00FF_FFFF address) is capable of encoding it on 3 bytes. This, being given that `DbAddress` now points to storage and that 3 bytes allow to address up to 64GB, should save some bytes for mapping
2. smaller memory setup for various tests


Scooletz:
No longer applicable to how the storage is persisted with Paprika. Now, with 1,000,000 fanout at the top it's much different.

PR #320, opened 2024-04-24
Fix FanOutPage new created page type

benaadams:
Currently marks it as a FanOutPage rather than a DataPage


Scooletz:
Closing, not needed anymore as there's no fanout atm. Also, page types changed a lot.

PR #328, opened 2024-05-10
DataPage.TryFlushDownToExistingChildren should delete on flushing down

Scooletz:
This PR fixes the flush down behavior for child `DataPage`s. If a value is pushed down, it should always be deleted. This drastically speeds up the execution (no double loop over map in DataPage) and potentially saves some space in there.

Test is needed.


Scooletz:
Merging without the test. There should be some better way of stating whether a page contains something or not so that the test can be more like property based testing instead of an imperative approach.

PR #331, opened 2024-05-14
Finalized blocks can be squashed

Scooletz:
This PR introduces a capability of squashing blocks which are finalized, if the queue of the finalized blocks is long enough. This should help in two cases:

1. Importing, with the queue size set to at least `blockchain.BlockCountForMaxCatchingUp`
2. Node, catching up with blocks that are applied one after another


Scooletz:
Replaced by #333

PR #339, opened 2024-05-28
Big Root Page

Scooletz:
This PR makes the `RootPage` much bigger, allowing for a broader fan out at the top (changed from 2 nibbles to 3 nibbles now). This should help with addressing the long application time for blocks. Similar to #338 it's based on Nethermind client having hard time to apply blocks faster then they are processed.

### Benchmarks

image

Sadly, the result db is bigger than original

image


Scooletz:
Closing as there are no gains visible here.

PR #341, opened 2024-05-30
Squeeze more bytes out of slottedarray

AryanGodara:
WIP

The current plan is to :-

1. Add separate method to check for mid length nibblepaths (4,5,6)
2. Update `long` nibblepaths from >=5 earlier to >=7 now.
3. Use 3 nibbles (instead of current 4) of the hash for storing the nibbles, and then use the fourth nibble to store the length marker (1,2,3 OR 4,5,6 OR beyond >=7). [effectively reducing 1 nibble from each slot?]
4. Update WriteImpl in NibblePath.cs, as currently first byte is used to store (length << 1 | oddity) , instead for length 4,5,6 since we already store the length in the preamble, we can directly store the trimmed nibblepath.
5. Update the PrepareKey UnprepareKey functions to add this intermediary case, where it should fetch the trimmed nibblepath differently now: 
    1) No trimmed for <=3, 
    2) Trimmed with length in preamble for 4,5,6 
    3) Longer keys, where 1st byte of trimmed path has length & oddity metadata

And some other details I haven't mentioned here, but I'm not able to coherently write them down today.
Only pushing the first commit to start the PR, will push the rest tomorrow, after going over once again and mapping the changes properly.

What the scope of this PR doesn't include: "introducing a special-case to make SlottedArray aware of the most common key at the given level."

### Assumptions (need to discuss further)

My assumption for point: "putting the length for keys of length 1-3 as the first nibble in hash, leaving 4+ to be encoded in the preamble"

length 0: special empty
length 1,2,3:  (2 bits) go in the hash.
for preamble, we have 3 bits active (
000 -> mask for: all nibble path (upto length 3) was in the hash itself.
111 -> mask for: len >= 10

Remaining: 001, 010, 011, 100, 101, 110 -> upto length 9 may be entered directly in the preamble.

So, for length >=10 we store length in the trimmed nibblepath instead of the preamble. And for the lesser ones, we saved 1 byte of space.
We already were saving that 1 byte of space for length <= 4 initially,
but if we only use 3 nibbles and only store upto length 3 in hash entirely, we save another nibble from the hash in each slot.

## Benchmark Comparison :-
```
New Benchmarks (My Branch):-
| Method                              | sliceFrom | Mean         | Error      | StdDev     |
|------------------------------------ |---------- |-------------:|-----------:|-----------:|
| Write_whole_page_of_data            | ?         | 20,775.28 ns | 113.080 ns |  88.285 ns |
[...]

PR #347, opened 2024-06-07
SlottedArrayBenchmarks.Hash_collisions uses a fixed size to make it comparable

Scooletz:
This PR changes one of the benchmarks to make it fixed size, so that the latter optimizations are comparable (see: #341 for example). It also sets the assertion so only the values that are set are searched for (no looping over various keys).


AryanGodara (approved):
LGTM 😎 
Ready for merge

PR #354, opened 2024-06-13
Merkle nodes with even-length paths made shorter

Scooletz:
This PR amends the way `Leaf`s of Merkle trie with even paths are stored, so that if the path starts with a nibble in form of `0b11__` it will be one byte shorter to store. As `0b11` is a one of four prefixes, with a uniform distribution 1/4 of even paths should require one byte less to store.

The optimization uses the fact that `Type.Leaf` is encoded in first byte as `0b1000_0000`. This plus the fact that other higher bits are not used by other cases, allows to use `0b11` as a discriminator of a leaf that has some property. In that case it's the property that the path starts with `0b11` which allows us to not store any additional data beside the leaf's `NibblePath`. The leaf type is encoded in the path itself.

### Benchmark

`Paprika.Cli` stats for `Sepolia` db.

#### Before 

image

### DIFF
--- a/src/Paprika.Tests/Merkle/NodeTests.cs
+++ b/src/Paprika.Tests/Merkle/NodeTests.cs
@@ -176,9 +176,13 @@ public class NodeTests
         decoded.Equals(leaf).Should().BeTrue($"Expected {leaf.ToString()}, got {decoded.ToString()}");
     }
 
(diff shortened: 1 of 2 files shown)


Scooletz:
Little to no gains.

PR #357, opened 2024-06-19
IFlushStrategy introduced to enable a catch up behavior

Scooletz:
This PR separates the decision of whether to flush or not to a separate component `IFlushStrategy`. It also removes the min timespan to flush as now it should spin fast, depending on the strategy, and only then, when the channel is empty, flush it (depending on the level)


Scooletz:
Discontinued. Blocks should be applicable in a faster way with regular application, not hacks like this.

PR #360, opened 2024-06-21
Different prefetcher

Scooletz:
This PR alters how the Merkle `prefetcher` works. It's different from the previous one in the following way.

1. Prefetcher is usable as is, requiring no additional work in Nethermind state.
1. All prefetches are scheduled separately with ` ThreadPool.QueueUserWorkItem`
1. The scheduling call doesn't do the mapping of `Address` ->`Keccak` nor `StorageCell`->`(Keccak, Keccak)`. It checks whether to run a frefetch using BitFilter with hash of raw inputs and then scheduled the work item
1. To allow the check whether something was run or not, a `IPrefetcherMapping` is introduced that allows to pass an arbitrary parameter (Address, StorageCell) and define the mapping that will be run in the thread pool work item.

### Benchmarks

Using `PrefetcherTests.Spin` scenarios. _Accounts + storage_ means that each account has 4 slots occupied, so at least one branch will be there.

| Scenario | Prefetcher used | Total commit time |
| --- | --- | --- |
| Only accounts  | n  | 13s |
| Only accounts  | y  | 11s |
| Accounts + storage | n | 30s | 
| Accounts + storage | y | 24s |

### DIFF
--- a/src/Paprika.Tests/Chain/PrefetchingTests.cs
+++ b/src/Paprika.Tests/Chain/PrefetchingTests.cs
@@ -55,7 +55,7 @@ public class PrefetchingTests
             db.AllowAllReads();
 
             // prefetch first
-            var p = block.OpenPrefetcher();
+            var p = block.OpenPrefetcher<IdentityPrefetchMapping, Keccak, Keccak>();
             p!.CanPrefetchFurther.Should().BeTrue();
             p.PrefetchAccount(keccak);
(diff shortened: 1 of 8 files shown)


Scooletz:
Something does not work. It causes invalid blocks on Sepolia.

image

Scooletz:
Closed in favor of #424

PR #362, opened 2024-06-24
Create static subclasses for WithPreamble and NoPreamble approaches for SlottedArray

AryanGodara:


Scooletz:
This is being substituted by the `KeyEncoding` provided in #375

PR #365, opened 2024-06-27
Batching for blocks enabled

Scooletz:
A PR that allows to batch blocks before applying them to the database if the number of blocks in the queue is sufficiently big that it breaches the history depth. If the number is bigger, blocks can be squashed before applying them to the database so that there's less book keeping in regards to the copying data in the database.


Scooletz:
This is not the crux of the performance issue. Blocks should be applicable much faster without any batching needed.

PR #369, opened 2024-07-02
SlottedArray faster search

Scooletz:
This PR augments `SlottedArray` to make it faster so that the data application is faster as well.

1. 4c0a052 - `TryGetImpl` changed from `bool` returning to index returning results in shorter code, from `809` to ` 777`
2. 70e765d - removed not needed `IsDeleted` check `; Total bytes of code 738`


Scooletz:
Will be replaced by #370 and a custom augmentation of the search

PR #370, opened 2024-07-04
Slotted array search and insert optimizations

Scooletz:
This PR tries to address the crux of the db performance issues, `SlottedArray` component that is visible in both, inserts and gets from the database. It does it in the following way:

1.  `TryGetImpl` changed from bool returning to index returning results in shorter code (~40 bytes)
2.  `IsDeleted` checkis not needed and can be removed saving one bool check and reducing the size further
3. `MarkAsDeleted`, when marking the entry as deleted changes its hash (to simple ~hash) so that it will no longer hit the search. See the `Set_And_Delete` benchmark below. This is 💥  for both sets and gets as they use this function heavily. The gains are tremendous.
4. Benchmarks showed that aligning the search to a full vector size is too heavy and does not bring gains to the search. For example, a test was done to align 7 items to get them to 8, or 15 to 16. No visible gains.

### Benchmarks

Selected benchmarks used to check the speed.

#### Before

| Method         | Mean     | Error    | StdDev   | Code Size |
|--------------- |---------:|---------:|---------:|----------:|
| Set_And_Delete | 48.63 us | 0.504 us | 0.447 us |   5,888 B |

#### After

| Method         | Mean     | Error     | StdDev    | Code Size |
|--------------- |---------:|----------:|----------:|----------:|
| Set_And_Delete | 6.005 us | 0.0482 us | 0.0451 us |   5,879 B |


src/Paprika/Data/SlottedArray.cs
```diff
@@ -62,7 +62,8 @@ public bool TrySet(in NibblePath key, ReadOnlySpan<byte> data)
 
     private bool TrySetImpl(ushort hash, byte preamble, in NibblePath trimmed, ReadOnlySpan<byte> data)
     {
-        if (TryGetImpl(trimmed, hash, preamble, out var existingData, out var index))
+        var index = TryGetImpl(trimmed, hash, preamble, out var existingData);
```
LukaszRozmej, line 65: maybe better that it would return a `bool` and `index` would be out param?

Scooletz: This was done before. As https://github.com/NethermindEth/Paprika/pull/369 showed it cuts off 40 bytes from 800 byte long method without impact on the call sites that need to compare it anyway.

PR #375, opened 2024-07-16
New fanout layout

Scooletz:
This PR reintroduces a few breaking features that makes the queries shorter and the db smaller.

### Big fan-out

The new layout for with much bigger fan out for `Storage`. It does it by encoding address lists effectively, allowing to pack over 1024 addresses in a single 4kb page. Then, it uses 2 layers of such structure to provide a total of 1 million buckets for `Storage`. The same mechanism is applied for contract `Ids` with a slightly smaller spread.

### Better key encoding

`SlottedArray` gains much better encoding for shorter keys. If a `NibblePath` is of length up to 4, the old mechanism ensured that no key bytes are written in the payload. For 5 and 6 it was 2 bytes. Now, the encoding can use 1 byte less for the key of length 5 and potentially (75% of cases) for the key of length 6. If you take into consideration that with each level, `Merkle` keys are shorter and prone to take a few nibbles, this could save a lot. In regards to the data size, `Leaf`s were already optimized in #293 and `Branch`es elsewhere so it's the keys now. For actual storage cells, that are 32 + 4 bytes at the top... it won't make the difference. These values don't occupy too much space anyway.

### No recursion on setting values

`DataPage` has a brand new `Set` method that is not recursive and won't be blowing up the stack. The recursion was replaced by a loop and now the stack variables are reused. The only case that is different is the `LeafLeftover` that now is also handled by the `DataPage`. This allows to minimize the number of page types and keep the implementation tight. For now, only one leftover is allowed, but this can be extended after tests.

### Write-through

The write-through behavior, greatly reduces the bulk insertion time. Once the database notices that there are a lot of keys with a given prefix (a child of a `DataPage` under the given nibble was written), it will pass through the writes, deleting previous versions alongside. This reduced time to apply block greatly.

### Allow more aggressive page reuse

If a page was written in the given batch and is registered for a reuse, it should be reused as soon as possible. This is done by adding one parameter to a registration and holding a stack of suck pages for getting them as soon as possible.

### Visitor based stats

The reporting API was a duplication of visitor. Now, with the prefix calculated properly, visitor is all is needed to go through the tree and this is used to report the data.

image

### Minor refactors

Additional minor refactors:

1. The decision what to flush down is still based on counting, but it's extracted and easy to be changed to size-based.

### Related

The following were an inspiration for this PR:

1. #366 so that 4 first nibbles are a subject of fanout. This, with higher capacity to amortize writes should help a lot.
3. #367 so that Merkle pages write through if a child was updated in the same batch


Scooletz:
Shift to #380 once its done

Scooletz:
Midimport stats

image

Scooletz:
Stats from importing 100 million of account of 260 million of mainnet

image

PR #382, opened 2024-08-22
Delete by prefix delete

Scooletz:
This PR introduces a new top-level method `IBatch.DeleteByPrefix (in Key key)` that allows a deletion of an arbitrary subtree pointed to by the key.

### DIFF
--- a/src/Paprika/Store/RootPage.cs
+++ b/src/Paprika/Store/RootPage.cs
@@ -165,7 +165,9 @@ public readonly unsafe struct RootPage(Page root) : IPage
     }
 
     private static uint ReadId(ReadOnlySpan<byte> id) => BinaryPrimitives.ReadUInt32LittleEndian(id);
-    private static void WriteId(Span<byte> idSpan, uint cachedId) => BinaryPrimitives.WriteUInt32LittleEndian(idSpan, cachedId);
+
+    private static void WriteId(Span<byte> idSpan, uint cachedId) =>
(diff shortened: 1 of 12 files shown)


src/Paprika/Store/RootPage.cs
```diff
@@ -225,12 +230,49 @@ public void Destroy(IBatchContext batch, in NibblePath account)
...
+            var data = batch.TryGetPageAlloc(ref Data.StateRoot, PageType.Standard);
+            var updated = new Merkle.StateRootPage(data).DeleteByPrefix(prefix.Path, batch);
+            Data.StateRoot = batch.GetAddress(updated);
+        }
+        else
+        {
+            scoped NibblePath id;
+            Span<byte> idSpan = stackalloc byte[sizeof(uint)];
+
+            var keccak = prefix.Path.UnsafeAsKeccak;
```
damian-orzechowski, line 248: Is it ok to use `UnsafeAsKeccak` here for paths that will be < 64 ?

damian-orzechowski (approved):

src/Paprika/Store/RootPage.cs
Scooletz: In line 237, there's a `.IsState` check. If the path is shorter than 64 or is 64 and is account, it's a state tree. It'll never bee less than 64 for the storage. The Merkle, the root of a storage tree will have the path set to 64.

PR #387, opened 2024-09-09
Linked overflow

Scooletz:
This PR changes how overflow pages work. Instead of having a complex logic of splitting keys by the first nibble (a part of it, masked to match the number of buckets), a simple linked list is used. This introduces some overhead for finding a value as all the elements in the list must be traversed, but can greatly decrease the amount of pages used for storing data.

Currently, a single `LeafOverflowPage` will point only to another one, making a chain of length of 2. This already shows nice size reduction. If generalized, this can be a succinct implementation with one or maybe two levels more.


Scooletz:
image

Scooletz:
Closed in favor of  #405 that if modified to use the linked list, can be make it a bit better by separating the type of the page.

Issue #265, opened 2024-03-01
Reduce disk bloat on Windows


Scooletz:

shubham-sonthalia:
Hi @Scooletz, are we focusing on PagedDB block (which uses the memory-mapped files) here?

Scooletz:
Yes sir! @shubham-sonthalia

shubham-sonthalia:
At the below line, we are initializing the memory mapped file. Do we have to come up with a way to divide the max_size of the Paprika DB into equal partitions (like 5 chunks) and start with 1/5th of the max DB space? or we want it to be dynamic?

https://github.com/shubham-sonthalia/Paprika/blob/666f37225fa14e1adf0d6970f1d911aa8b7b1a49/src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs#L44

<img width="596" alt="image" src="https://github.com/user-attachments/assets/d7c7d7fc-7e49-4b29-9f97-2d67fb8d7a0d">

Scooletz:
We could be dynamic, but dynamism can come with a price. If we map it separately, there will be different pointers for example.

B1n4ryR41nb0w:
Can I please get it assigned, if that's possible?

PR #388, opened 2024-09-10
#265 Update MemoryMappedPageManager memory mapped initiation

shubham-sonthalia:
#265 

Two changes: 

1. Creating the MemoryMappedFile without setting the total file size upfront 

_mapped = MemoryMappedFile.CreateFromFile(_file, null, 0, MemoryMappedFileAccess.ReadWrite, HandleInheritability.None, true);

2. Instead of mapping the whole file, mapping only a portion of it dynamically as needed

_whole = _mapped.CreateViewAccessor(0, historyDepth * Page.PageSize);

### DIFF
--- a/src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
+++ b/src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
@@ -61,11 +61,9 @@ public sealed class MemoryMappedPageManager : PointerPageManager
             _file = new FileStream(Path, FileMode.Open, FileAccess.ReadWrite, FileShare.None, 4096,
                 PaprikaFileOptions);
         }
+        _mapped = MemoryMappedFile.CreateFromFile(_file, null, 0, MemoryMappedFileAccess.ReadWrite, HandleInheritability.None, true);
 
-        _mapped = MemoryMappedFile.CreateFromFile(_file, null, (long)size, MemoryMappedFileAccess.ReadWrite,
-            HandleInheritability.None, true);
-
-        _whole = _mapped.CreateViewAccessor();
+        _whole = _mapped.CreateViewAccessor(0, historyDepth * Page.PageSize);
         _whole.SafeMemoryMappedViewHandle.AcquirePointer(ref _ptr);
         _options = options;
 


shubham-sonthalia:
Hi @Scooletz, pls let me know if these changes are in the right direction.

src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
```diff
@@ -61,11 +61,9 @@ public unsafe MemoryMappedPageManager(long size, byte historyDepth, string dir,
             _file = new FileStream(Path, FileMode.Open, FileAccess.ReadWrite, FileShare.None, 4096,
                 PaprikaFileOptions);
         }
+        _mapped = MemoryMappedFile.CreateFromFile(_file, null, 0, MemoryMappedFileAccess.ReadWrite, HandleInheritability.None, true);
 
-        _mapped = MemoryMappedFile.CreateFromFile(_file, null, (long)size, MemoryMappedFileAccess.ReadWrite,
-            HandleInheritability.None, true);
-
-        _whole = _mapped.CreateViewAccessor();
+        _whole = _mapped.CreateViewAccessor(0, historyDepth * Page.PageSize);
```
Scooletz, line 66: The `_whole` is used as both, the view to `ForceFlush` when a synchronization happens and as an accessor that provides the pointer `_ptr` for the whole file. This is why it was set to cover whole file.

src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
```diff
@@ -61,11 +61,9 @@ public unsafe MemoryMappedPageManager(long size, byte historyDepth, string dir,
             _file = new FileStream(Path, FileMode.Open, FileAccess.ReadWrite, FileShare.None, 4096,
                 PaprikaFileOptions);
         }
+        _mapped = MemoryMappedFile.CreateFromFile(_file, null, 0, MemoryMappedFileAccess.ReadWrite, HandleInheritability.None, true);
```
Scooletz, line 64: AFAIK, the mapping will get size from the `_file`, but if the file is just created, it won't work as expected as one will need to set the length. I don't understand how this should work.

Scooletz (commented):
Provided comments. I don't understand how this could work without significant changes or using the NT method mentioned in the referred issue. Maybe we should add some tests for the `MemoryMappedPageManager` to ensure that it works properly? Also, add some matrix for OS dependent tests?

shubham-sonthalia:
> Provided comments. I don't understand how this could work without significant changes or using the NT method mentioned in the referred issue. Maybe we should add some tests for the `MemoryMappedPageManager` to ensure that it works properly? Also, add some matrix for OS dependent tests?

I read that the .NET memory-mapped file API abstracts the complexity of memory management, so we don’t need to use lower-level functions like NtCreateSection in C#.

src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
shubham-sonthalia: what if we remove this line - 

_file.setLength(size);

while creating a new file. We can set it to some safe initial size.

src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
shubham-sonthalia: Can we introduce a new variable called _viewAccessor for accessing the memory mapped portion as a pointer. And use _whole as it is for `ForceFlush`?

Scooletz: I'm not following. `ForceFlush` semantics should be that all the pages that were allocated are flushed down and then FSYNC is called. I believe we could create a one time use accessor and flush it. This might work.

src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
Scooletz: I don't know how to answer. We may do all the things as long as the file is properly mapped and written and flushed 😅

shubham-sonthalia: Would you recommend - 

1. initialize the _file with an initial size of say, 4 MB. 
2. modifying the write and flush operations in a way it first checks the size of the _file and only if it's large enough to store incoming data, proceed as it is. Otherwise, can double the file size until it can accommodate incoming data.

src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
shubham-sonthalia: You mean, inside the ForceFlush method, we can get another accessor for _file that covers the whole file and flush it?

Scooletz:
Any comments @shubham-sonthalia ?

src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
shubham-sonthalia: > Would you recommend -
> 
> 1. initialize the _file with an initial size of say, 4 MB.
> 2. modifying the write and flush operations in a way it first checks the size of the _file and only if it's large enough to store incoming data, proceed as it is. Otherwise, can double the file size until it can accommodate incoming data.

@Scooletz  What do you think about this approach?

PR #391, opened 2024-09-13
Fix encoding for branch and extension for inlined data.

damian-orzechowski:
Fix RLP encoding for small inlined data:
- Branch - when slicing target span to write correct sequence length, the target position is leaving extra 0 byte for content length < 56
- Extension - should write child's `Keccak` or `Rlp` - `contentLength` already takes that into account

### DIFF
--- a/src/Paprika/Merkle/ComputeMerkleBehavior.cs
+++ b/src/Paprika/Merkle/ComputeMerkleBehavior.cs
@@ -649,7 +649,9 @@ public class ComputeMerkleBehavior : IPreCommitBehavior, IDisposable
         // Write length of length in front of the payload, resetting the stream properly
         var end = stream.Position;
         var actualLength = end - initialShift;
-        var lengthOfLength = Rlp.LengthOfLength(actualLength) + 1;
+        var lengthOfLength = Rlp.LengthOfLength(actualLength);
+        if (actualLength >= 56) //to match StartSequence
+            lengthOfLength++;
         var from = initialShift - lengthOfLength;
         stream.Position = from;
         stream.StartSequence(actualLength);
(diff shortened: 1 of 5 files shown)


src/Paprika/Merkle/ComputeMerkleBehavior.cs
```diff
@@ -649,7 +649,9 @@ private void EncodeBranch(scoped in Key key, scoped in ComputeContext ctx, scope
         // Write length of length in front of the payload, resetting the stream properly
         var end = stream.Position;
         var actualLength = end - initialShift;
-        var lengthOfLength = Rlp.LengthOfLength(actualLength) + 1;
+        var lengthOfLength = Rlp.LengthOfLength(actualLength);
+        if (actualLength >= 56) //to match StartSequence
```
Scooletz, line 653: Should we have it extracted as a const under `Rlp` maybe? Then both `Rlp.LengthOfLength` and this place could refer to it properly?

src/Paprika/Merkle/ComputeMerkleBehavior.cs
```diff
@@ -692,7 +694,10 @@ private void EncodeExtension(scoped in Key key, scoped in ComputeContext ctx, sc
         RlpStream stream = new(pooled.Span.Slice(slice, totalLength));
         stream.StartSequence(contentLength);
         stream.Encode(span);
-        stream.Encode(keccakOrRlp.Keccak);
+        if (keccakOrRlp.DataType == KeccakOrRlp.Type.Rlp)
```
Scooletz, line 697: Clearly, after following the test, this check was missing.

Scooletz (changes requested):
One change request. The fix itself: 👍

src/Paprika/Merkle/ComputeMerkleBehavior.cs
damian-orzechowski: Added constant in `Rlp`

Scooletz (approved):

PR #393, opened 2024-09-18
#384 Initial Changes for confirmation

### LINKED ISSUE #384: Delete accounts by prefix

shubham-sonthalia:
Resolves #384 

List of Changes - 

1. A new property in IBatchContext called as NullPage set to new Page((*ptr)0). 
2. All DeleteByPrefix methods checking if prefix is empty, then registering the page for future reuse and returning batch.NullPage. 

Pending Changes - 
Recursively calling for child pages (if the prefix is empty) - need help here.

### DIFF
--- a/src/Paprika/Store/IBatchContext.cs
+++ b/src/Paprika/Store/IBatchContext.cs
@@ -72,6 +72,8 @@ public interface IBatchContext : IReadOnlyBatchContext
     Page TryGetPageAlloc(ref DbAddress addr, PageType pageType);
 
     BatchStats? Stats { get; }
+
+    public unsafe Page NullPage => new Page((byte*)0);
 }
 
 public class BatchStats : IBatchStats
--- a/src/Paprika/Store/BatchContextBase.cs
+++ b/src/Paprika/Store/BatchContextBase.cs
@@ -39,7 +39,7 @@ abstract class BatchContextBase(uint batchId) : IBatchContext
 
     public bool WasWritten(DbAddress addr) => GetAt(addr).Header.BatchId == BatchId;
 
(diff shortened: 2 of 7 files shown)


shubham-sonthalia:
Hey @Scooletz do these changes look good?

src/Paprika/Store/IBatchContext.cs
```diff
@@ -72,6 +72,8 @@ Page EnsureWritableCopy(ref DbAddress addr)
     Page TryGetPageAlloc(ref DbAddress addr, PageType pageType);
 
     BatchStats? Stats { get; }
+
+    public unsafe Page NullPage => new Page((byte*)0);
```
Scooletz, line 76: If we pass null pointer here, some things may blow up. Which might be ok, but I think that the preffered way of doing it, would be to have something that can be naturally retrieved by `.GetAddress`. Having that said, I'd go with

```suggestion
    public unsafe Page NullPage => new Page(GetAt(DbAddress.Null));
```

Scooletz (changes requested):
Looks really promising 🙂 
Some test could be added to ensure that the tree is properly rolled up and the pages are reused. Like: writing a big tree in one block, removing in other, spinning one block to make them reusable and reuse again.


Scheduled a run for GH workflow and added some remarks that should improve the logic.

ssonthal:
Hi @Scooletz , i made the suggested fix in the NullPage declaration. Will work on the test cases like you suggested.

ssonthal:
Hey @Scooletz , I couldn't pick it up earlier due to other work. Resuming it now. Can you kindly help me with existing code references to write these test cases?

I looked over the PagedDbTests.cs and existing DeleteByPrefix test methods. But I think what you described requires a different approach.

PR #396, opened 2024-09-26
Platform specific prefetcher

Scooletz:
This PR migrates the prefetching from SSE based to an OS specific solution. The prefetch happens on a separate thread that accumulates pages to be prefetched in a channel and schedules them in bulks. Similar implementation could be provided for not needing pages anymore, but it would require either nop implementation on Windows or a better understanding of advising on Win. `DiscardVirtualMemory` cannot be used as it destroys the content which is not what is needed here. What is needed is a hint that the given area won't be used anymore.

The following methods are used to hint that given addresses will be used:

1. `PrefetchVirtualMemory` on `Windows` 
2. `madvise` on `Linux`
3. on MacOS, currently not implemented

While testing stats gathering, issuing prefetches when `.Accept` on a page is called increased the IO throughput significantly which can be show by running `cli stats`. 

#### Before

image

#### After

image


src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
```diff
@@ -80,6 +97,51 @@ public unsafe MemoryMappedPageManager(long size, byte historyDepth, string dir,
...
+        }
+    }
+
+    private async Task RunPrefetcher()
+    {
+        var reader = _prefetches.Reader;
+
+        while (await reader.WaitToReadAsync())
+        {
+            PrefetchImpl(reader, this);
```
dipkakwani, line 119: A basic question: Since the prefetcher runner thread is running asynchronously, is it theoretically possible that for a certain address it executes the prefetch **_after_** the actual read on that address is already done? Prefetcher will be able to find that the address is already in memory, but still these calls can effectively double the number of IOs?

src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
```diff
@@ -24,6 +27,18 @@ public sealed class MemoryMappedPageManager : PointerPageManager
     private readonly MemoryMappedViewAccessor _whole;
     private readonly unsafe byte* _ptr;
 
+    // Prefetcher
+    private const int PrefetcherCapacity = 1000;
```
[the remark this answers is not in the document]
dipkakwani: Also, could that configuration parameter be used to turn-off prefetching altogether in case it is not required?

dipkakwani (approved):

src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
```diff
@@ -24,6 +27,18 @@ public sealed class MemoryMappedPageManager : PointerPageManager
     private readonly MemoryMappedViewAccessor _whole;
     private readonly unsafe byte* _ptr;
 
+    // Prefetcher
```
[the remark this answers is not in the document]
Scooletz: Ah, it could be made more specific as this is about the `Sequential` flag as we don't do `Sequential` because clearly it's a random access all over the place. Will amend.

src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
Scooletz: It is possible that it will run after the page is fetched. And yes, it's possible that it will be fetching over pages that are already there. We could then discuss whether to leave SSE-based soft prefetch and leave this implementation for heavier use. 

The place that I'd still see it as is, would be the `AbandonedList` that prepares the next page that will be copied to. In this case, the page is returned, then copied to etc while the next is scheduled for a prefetch. This should give it enough time. 

The next place that it works (just following the raw execution tiem) is visiting. We schedule prefetching for all the children up front and as it walks in the preorder, beside the first child the rest most likely will have time to be prefetched. Again, raw execution time shows a very nice boost (also, disk throughput is much higher)

The one place that makes it less easy to reason about is the `DataPage.TryGet`. Maybe, in this case we should use SSE prefetch. The page, if it faults, it will take more time anyway than a search across page (now ~10ns). So it might be the case that we'd like to skip issuing a costly OS call.

Scooletz:
@dipkakwani Summing up: maybe SSE prefetch shoudl be still available as a soft version and should be able to be used depending on the call site (`DataPage`)

src/Paprika/Store/PageManagers/MemoryMappedPageManager.cs
dipkakwani: Thanks for clarifying! Make sense to me to use two different variants based on the usage 😄

PR #403, opened 2024-10-09
Minor Optimizations in Abandoned Pages

dipkakwani:
This PR optimizes a few functions in abandoned pages and list.

### DIFF
--- a/src/Paprika/Store/AbandonedList.cs
+++ b/src/Paprika/Store/AbandonedList.cs
@@ -166,16 +166,9 @@ public struct AbandonedList
     {
         if (MaxCount == EntriesCount)
         {
-            // No place, find the youngest and attach to it
-            var maxAt = 0;
-
-            for (var i = 1; i < MaxCount; i++)
-            {
-                if (BatchIds[i] > BatchIds[maxAt])
-                {
-                    maxAt = i;
-                }
-            }
+            // No place, attach it to the youngest batch id
+            var maxAt = MaxCount - 1;
+            Debug.Assert(MaxCount > 2 && BatchIds[MaxCount - 1] > BatchIds[MaxCount - 2]);
 
             // 1. Attach the previously existing abandoned as tail to the current one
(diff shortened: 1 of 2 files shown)


src/Paprika/Store/AbandonedPage.cs
```diff
@@ -232,13 +233,13 @@ public bool TryPush(DbAddress address)
     public void AttachTail(DbAddress tail, IBatchContext batch)
     {
         Debug.Assert(page.Header.BatchId == batch.BatchId);
+        ref var abandoned = ref Data;
```
Scooletz, line 236: Why ref over `Data` here? Could it be `ref Data.Next` maybe?

Also the reference is not overwritten anywhere but assigned over and over. This will destroy the chain, will it not?

src/Paprika/Store/AbandonedList.cs
```diff
@@ -166,16 +166,9 @@ private void Register(DbAddress head, IBatchContext batch)
...
-
-            for (var i = 1; i < MaxCount; i++)
-            {
-                if (BatchIds[i] > BatchIds[maxAt])
-                {
-                    maxAt = i;
-                }
-            }
+            // No place, attach it to the youngest batch id
+            var maxAt = MaxCount - 1;
```
Scooletz, line 170: Nice. I'd add in a comment that it's sorted to last is the youngest.

Scooletz (commented):
Will this PR have tests amended? It looks at the moment that the tail chaining behavior is not covered as the first version looked faulty.

dipkakwani:
@Scooletz Yes, I'm trying to repro it with the older version. I ran all the existing tests under the debugger, and it seems that we never create a chain with more than 2 elements (data.next = null always), which is why even the first version of the implementation is passing. Trying to find out how to make it pack more elements and possibly a new simpler test to repro a failure.
Will cover that test in this PR.

dipkakwani:
Will work on the test in a separate PR as none of the existing code paths simulate re-inserting into an existing chain. Need a new test to simulate that behavior - discussed some ideas around that with Szymon.

PR #409, opened 2024-10-14
Add New Test for Abandoned Pages Chain

dipkakwani:
Follow up PR for https://github.com/NethermindEth/Paprika/pull/403 to improve the test coverage for abandoned pages. The new test (`Abandoned_chain_creation_with_overflow`) creates enough number of abandoned pages to form a chain and ensures that there are no leaks.

### DIFF
--- a/src/Paprika.Tests/Store/AbandonedTests.cs
+++ b/src/Paprika.Tests/Store/AbandonedTests.cs
@@ -247,7 +247,7 @@ public class AbandonedTests : BasePageTests
     [Test]
     public async Task Abandoned_chain_creation_with_overflow()
     {
-        // Minimum iterations required to overflow the abadoned list (AbandonedList.MaxCount).
+        // Minimum iterations required to overflow the abandoned list (AbandonedList.MaxCount).
         const int minIterations = 60;
 
         // Minimum accounts required to overflow an abandoned page (AbandonedPage.Payload.MaxCount * 2).
--- a/src/Paprika.Tests/Store/AbandonedTests.cs
+++ b/src/Paprika.Tests/Store/AbandonedTests.cs
@@ -245,6 +245,7 @@ public class AbandonedTests : BasePageTests
     }
 
     [Test]
+    [Category(Categories.LongRunning)]
     public async Task Abandoned_chain_creation_with_overflow()
     {
         // Minimum iterations required to overflow the abandoned list (AbandonedList.MaxCount).
@@ -259,7 +260,7 @@ public class AbandonedTests : BasePageTests
         var accountValue = new byte[2900];
         new Random(17).NextBytes(accountValue);
 
-        using var db = PagedDb.NativeMemoryDb(150000 * Page.PageSize, HistoryDepth);
+        using var db = PagedDb.NativeMemoryDb(140000 * Page.PageSize, HistoryDepth);
 
         // Start read only batch to ensure that new pages are allocated instead of reusing
         // the abandoned pages.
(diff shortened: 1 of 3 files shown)


src/Paprika.Tests/Store/AbandonedTests.cs
```diff
@@ -250,4 +243,61 @@ public async Task Reuse_in_grow_and_shrink()
             await block.Commit(CommitOptions.FlushDataAndRoot);
         }
     }
+
+    [Test]
+    [Category(Categories.LongRunning)]
+    public async Task Abandoned_chain_creation_with_overflow()
+    {
+        // Minimum iterations required to overflow the abandoned list (AbandonedList.MaxCount).
+        const int minIterations = 60;
```
Scooletz, line 252: This const is probably dependend on the `AbandonedList` max count, is it not? If it is, is should be taken from the `AbandonedList`. Otherwise, if we optimize the list and increase the bucket count it will provide false positive.

src/Paprika.Tests/Store/AbandonedTests.cs
```diff
@@ -250,4 +243,61 @@ public async Task Reuse_in_grow_and_shrink()
...
+
+        // Minimum accounts required to overflow an abandoned page (AbandonedPage.Payload.MaxCount * 2).
+        const int numAccounts = 2000;
```
Scooletz, line 255: Why not calculate the const on the basis of the comment here? Again, a change of implementation and consts should automatically change this test.

src/Paprika.Tests/Store/AbandonedTests.cs
```diff
@@ -250,4 +243,61 @@ public async Task Reuse_in_grow_and_shrink()
...
+
+        var keccaks = Initialize(numAccounts);
+
+        // Set big value.
+        var accountValue = new byte[2900];
+        new Random(17).NextBytes(accountValue);
+
+        using var db = PagedDb.NativeMemoryDb(140000 * Page.PageSize, HistoryDepth);
```
Scooletz, line 263: Less than 1GiB if I get this right. Is it heavy to have it runnable on CI and remove the `LongRunning` maybe? `LongRunning` are not run on CI atm.

src/Paprika.Tests/Store/AbandonedTests.cs
```diff
@@ -250,4 +243,61 @@ public async Task Reuse_in_grow_and_shrink()
...
+        // Try reusing the abandoned pages.
+        for (var i = 0; i < minIterations; i++)
+        {
+            using var block = db.BeginNextBatch();
+            foreach (var keccak in keccaks)
+            {
+                block.SetAccount(keccak, accountValue);
+            }
+
+            block.VerifyDbPagesOnCommit();
```
Scooletz, line 291: Hmmm, there's no actual assert here on the structure of Abandoned pages. Should we update visitor maybe and allow to walk through the list? Or maybe, externalize the list somehow from the root and allow to inspect it here to assert chaining?

dipkakwani: Need to think more about this one, I don't see any easy implementation without disrupting the existing code. We can assert on at least that abandoned list is full, but to fetch the exact abandoned pages and chaining, the complete Address structure has to be exposed either to Root or to the test directly.

dipkakwani: Missing visitor actually did catch the issue of non-reachability of pages with incorrect chain implementation, so it does provide some assertion.

Scooletz: Ah, I see. Then I agree. If we're good from the reachability pov it provides some test for not dropping the tail.

Scooletz:
@dipkakwani One conflict to solve

Scooletz (approved):

Issue #410, opened 2024-10-17
Optimize AbandonedList space usage


Scooletz:
Currently, `AbandonedList` uses two spans to keep the data. One for the batch ids and the other for addresses of the abandoned pages. As each `AbandonedPage` keeps the batch in itself, the storage for batch ids is not needed and can be removed safely. Whenever a batch id of the given `AbandonedPage` is needed, it can be retrieved from the page itself. This would double the number of entries in `AbandonedList` and make it even less likely to create chains of pages.

https://github.com/NethermindEth/Paprika/blob/3dea5e4a75248c6f9ec5e4f2493f3ca1913355c7/src/Paprika/Store/AbandonedList.cs#L25-L32

PR #412, opened 2024-10-17
Optimize AbandonedList Space Usage

dipkakwani:
Remove BatchIds from the Abandoned List as it is redundant information which is already stored in each Abandoned page. https://github.com/NethermindEth/Paprika/issues/410

### DIFF
--- a/src/Paprika/Store/AbandonedList.cs
+++ b/src/Paprika/Store/AbandonedList.cs
@@ -11,7 +11,7 @@ namespace Paprika.Store;
 public struct AbandonedList
 {
     /// <summary>
-    /// The start for spans of <see cref="BatchIds"/> and <see cref="Addresses"/>.
+    /// The start for <see cref="Addresses"/> span.
     /// </summary>
     private const int EntriesStart = DbAddress.Size + sizeof(uint);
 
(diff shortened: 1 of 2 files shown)


src/Paprika/Store/AbandonedList.cs
```diff
@@ -166,15 +158,13 @@ private void Register(DbAddress head, IBatchContext batch)
     {
         if (MaxCount == EntriesCount)
         {
-            // No place, attach it to the youngest batch id from the sorted BatchIds
+            // No place, attach it to the last page which has the youngest batch id. Addresses
+            // are always sorted by batch ids, hence the last page.
             var maxAt = MaxCount - 1;
-            Debug.Assert(MaxCount > 2 && BatchIds[MaxCount - 1] > BatchIds[MaxCount - 2]);
```
dipkakwani, line 171: Removed the debug assert to keep the code shorter. We can fetch the corresponding batch ids from the pages, but not sure if it is required?

Scooletz: I think it's ok now.

Scooletz (changes requested):
One thing to fix. 
Also, maybe a test for it? There was the chaining one, why not to use it again?

src/Paprika/Store/AbandonedList.cs
```diff
@@ -22,10 +22,6 @@ public struct AbandonedList
     [FieldOffset(0)] private DbAddress Current;
     [FieldOffset(DbAddress.Size)] private uint EntriesCount;
 
-    [FieldOffset(EntriesStart)] private uint BatchIdStart;
```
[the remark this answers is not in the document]
dipkakwani: My bad, completely missed changing the offsets 😓 
Fixed it in the latest iteration and also added a sanity check on the MaxCount value (without exposing complete data structure details).

Scooletz (approved):
😍

PR #415, opened 2024-10-18
Delete by short prefix on state root page

damian-orzechowski:
Remove restriction of removing data from root page with a prefix shorter than `ConsumedNibbles` (2). Highly unlikely in real world scenarios, but greatly improves testing (especially fuzzing with smaller trie sizes).
Depends on #414.

### DIFF
--- a/src/Paprika/Store/StateRootPage.cs
+++ b/src/Paprika/Store/StateRootPage.cs
@@ -74,7 +74,27 @@ public readonly unsafe struct StateRootPage(Page page) : IPage
             return new StateRootPage(writable).DeleteByPrefix(prefix, batch);
         }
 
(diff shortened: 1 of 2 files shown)


src/Paprika/Store/StateRootPage.cs
```diff
@@ -74,7 +74,27 @@ public Page DeleteByPrefix(in NibblePath prefix, IBatchContext batch)
...
+            {
+                var idx = prefix.Nibble0 << NibblePath.NibbleShift;
+                for (int i = 0; i < 16; i++)
+                {
+                    ref var addrShort = ref DeleteAll(Data.Buckets, idx + i);
+                }
+            }
+            else if (prefix.IsEmpty)
+            {
+                //can all pages be freed here?
```
Scooletz, line 89: I think they can but this requires a separate PR. I think it should be up to the pages to register themselves and return a `null page` there.

PR #436, opened 2024-11-08
Unbalanced Parallel

benaadams:
Lower allocating Parallel.For that handles unbalanced work loads better

Rather than partitioning the range and each thread processing that number of items; it creates that number of ThreadPool worker (minus one for main thread); then processes the next available index until the range is complete, meaning one thread may process different numbers of items from another.

This should reduce overall end to end latency also.

Before

| Method                           | count | Mean      | Error    | StdDev   | Gen0   | Code Size | Allocated |
|--------------------------------- |------ |----------:|---------:|---------:|-------:|----------:|----------:|
| Or_BitMapFilter_OfN_128_Multiple | 16    |  24.82 us | 0.346 us | 0.289 us | 0.3662 |   2,464 B |   6.19 KB |
| Or_BitMapFilter_OfN_128_Multiple | 32    |  41.07 us | 0.807 us | 1.232 us | 0.3662 |   2,464 B |   6.21 KB |
| Or_BitMapFilter_OfN_128_Multiple | 64    | 221.41 us | 3.831 us | 2.991 us | 0.2441 |   2,447 B |   7.58 KB |

After

| Method                           | count | Mean      | Error    | StdDev   | Gen0   | Code Size | Allocated |
|--------------------------------- |------ |----------:|---------:|---------:|-------:|----------:|----------:|
| Or_BitMapFilter_OfN_128_Multiple | 16    |  27.71 us | 0.095 us | 0.079 us | 0.0610 |   1,126 B |   1.20 KB |
| Or_BitMapFilter_OfN_128_Multiple | 32    |  34.58 us | 0.557 us | 0.521 us | 0.0610 |   1,126 B |   1.45 KB |
| Or_BitMapFilter_OfN_128_Multiple | 64    | 180.98 us | 3.559 us | 6.139 us |      - |   1,126 B |   1.95 KB |


src/Paprika/Utils/WorkProcessor.cs
```diff
@@ -0,0 +1,181 @@
...
+                value = _data.Action(i, value);
+                i = _data.Index.GetNext();
+            }
+
+            _data.Finally(value);
+
+            _data.MarkThreadCompleted();
+        }
+
+        private class Data<TValue>(int threads, int fromInclusive, int toExclusive, Func<TValue> init,
```
Scooletz, line 153: Have you considered deriving from `Data` here? A lot is shared, beside the `local` related concepts?

src/Paprika/Utils/WorkProcessor.cs
```diff
@@ -0,0 +1,181 @@
...
+            ParallelOptions parallelOptions,
+            Func<TLocal> init,
+            Func<int, TLocal, TLocal> action,
+            Action<TLocal> @finally)
+        {
+            var threads = parallelOptions.MaxDegreeOfParallelism > 0 ? parallelOptions.MaxDegreeOfParallelism : Environment.ProcessorCount;
+
+            var data = new Data<TLocal>(threads, fromInclusive, toExclusive, init, action, @finally);
+
+            for (int i = 0; i < threads - 1; i++)
```
Scooletz, line 113: Same as below, have you considered extracting the loop and the wait condition? Not sure if duable to to the processor construction but maybe?

src/Paprika/Utils/WorkProcessor.cs
```diff
@@ -0,0 +1,181 @@
...
+            {
+                ThreadPool.UnsafeQueueUserWorkItem(new InitProcessor<TLocal>(data), preferLocal: false);
+            }
+
+            new InitProcessor<TLocal>(data).Execute();
+
+            if (data.ActiveThreads > 0)
```
Scooletz, line 120: Is it safe? What if the last thread decrements the threads, this one reads `ActiveThreads > 0` as it can be stale, and then locks on and waits without getting the pulse? Why not pulse N times or remove the check and perform active thread counting in the lock + pulse?

Scooletz (changes requested):
I think there might be an issue with the `Monitor.Wait`

src/Paprika/Utils/WorkProcessor.cs
benaadams: The lock is a memory barrier; so it shouldn't read stale inside

Either the lock is acquired here and the inner read is `> 0` so it waits, and the pulse will be sent

Or the pulse lock is acquired first so the ActiveThreads will be read as 0, and it will skip

src/Paprika/Utils/WorkProcessor.cs
benaadams: Moved to derived classes

Scooletz (approved):

PR #439, opened 2024-11-20
Vectorized Defragmentation

dipkakwani:
The existing de-fragmentation algorithm in SlottedArray scans all the slots sequentially and defragments the data structure by moving non-deleted slots together. This PR re-writes the algorithm with a new vectorized implementation.

The following benchmarks show the time taken for defragmentation with the old vs new algorithm:

image

image

Defragmentation - deleting first key followed by defragmentation
Defragmentation_Large - deleting `count` randomly selected keys followed by defragmentation


src/Paprika/Data/SlottedArray.cs
```diff
@@ -906,78 +910,198 @@ private void MarkAsDeleted(int index)
...
+
+         while (i < aligned)
+         {
+             if (i + jump < aligned)
+             {
+                 var slotData = Vector256.LoadUnsafe(ref d, (UIntPtr)i);
+
+                 // Ignore all the even entries from the vector as the preamble data is contained within
+                 // the first 2 bytes (ushort) out of the 4 bytes slot.
+                 slotData = Avx2.BlendVariable(slotData, Vector256<ushort>.Zero, Vector256.Create((ushort)0b_1010_1010_1010_1010));
```
Scooletz, line 956: Why is this needed? What does it do? Isn't it a bug? There's `BitwiseAnd` underneath so it should remove what is not needed? The comment states that the slot is 4 bytes long, but after #376 this is no longer the case in a sense that the preambles and hashes are grouped into vector sizes, see: https://github.com/NethermindEth/Paprika/blob/main/docs/design.md#slottedarray-layout

src/Paprika/Data/SlottedArray.cs
```diff
@@ -906,78 +910,198 @@ private void MarkAsDeleted(int index)
...
+                 
+                 var preambleData = Vector256.BitwiseAnd(slotData, preambleMask);
+                 
+                 if (Vector256.EqualsAny(preambleData, deletedMask))
```
Scooletz, line 960: I'd benchmark it. There was the case that xoring and asserting whether equals zero was a bit faster

src/Paprika/Data/SlottedArray.cs
```diff
@@ -906,78 +910,198 @@ private void MarkAsDeleted(int index)
...
+                 {
+                     // Some slots are deleted in this batch, process individually
+                     for (var j = 0; j < SlotsPerVector; j++)
```
Scooletz, line 963: This loop could be replaced. What we could do is to extract most significant bits after equality and move in larger chunks like we do in `TryFind`. Also, it copies slot by slot with no considerations for continuous deleted chunks?

src/Paprika/Data/SlottedArray.cs
```diff
@@ -906,78 +910,198 @@ private void MarkAsDeleted(int index)
...
+
+                         i += Slot.TotalSize;
+                     }
+                 }
+                 else
+                 {
+                     // No deleted slots in this batch, copy the whole batch
+                     if (i != writeTo)
+                     {
+                         CopyBatch(i, writeTo);
```
Scooletz, line 984: Ah, so we do batch copy but only if we find the whole batch valid. If there's a single delete we'll go one by one then?

src/Paprika/Data/SlottedArray.cs
```diff
@@ -906,78 +910,198 @@ private void MarkAsDeleted(int index)
...
+         }
+
+         // Adjust header values
+         _header.Low = (ushort)(writeTo * Slot.TotalSize);
+         _header.Deleted = 0;
+         RecalculateHigh();
+     }
+
+     // Helper function to copy a batch of slots
+     private void CopyBatch(int readFrom, int writeTo)
```
Scooletz, line 1022: This looks like a long method. I'd benchmark it heavily.

Scooletz (commented):
I think the main issue is with the masking of slots while it should not be done. Provided comments.

Scooletz:
One more general comment. Have you considered @dipkakwani splitting the search for deletions from the actual compaction of the array? Something that would mimic `TryGetImpl` to scan through the whole map and constructing a bit set of deletes and then act upon it? We do have an upper boundary on the number of entries (~1024 max for a whole page which gives 128 ulongs) so maybe search first, then move?

src/Paprika/Data/SlottedArray.cs
dipkakwani: Ah, I earlier assumed it to be 2 bytes and recently added this mask. I think I got misled by the total size parameter within Slot:

`public const int TotalSize = Size + sizeof(ushort);`

Just realized that it mentions in the comment that it's total size of hash and slot together, missed that completely. Maybe I will rename the parameter to `TotalSizeWithHash` for better clarity? From the current name TotalSize and Size look exactly similar :sweat_smile:

Scooletz:
> @Scooletz Yes that is an interesting point, I did think in this direction - but I wasn't sure if there is any advantage since eventually the elements have to be moved during defragmentation. The process to move once the deleted mask is formed would require moving all the elements between two deleted items in one shot; the number of these elements could be smaller, equal or longer than the vector size. I am not sure if there if there is any advantage with this move operation vs directly reading and moving vectors whenever possible.

Eventually you need to move elements, but if you had a set that marks entries that are alive with 1s and 0s where they are dead, you could potentially move in as big chunks as possible for both parts: moving `Slots` and moving actual payloads. This could greatly simplify the moving and also allow moving much bigger chunks in one go (less memcopy calls). There would some overhead in consuming this big bit map though. 🤔

src/Paprika.Benchmarks/SlottedArrayBenchmarks.cs
```diff
@@ -191,6 +227,43 @@ public int EnumerateNibble(byte nibble)
...
+    }
+    
+    [Benchmark]
+    [Arguments(1)]
+    [Arguments(25)]
+    [Arguments(50)]
+    [Arguments(96)]
+    public bool Defragmentation_Large(int count)
+    {
+        var map = new SlottedArray(new Span<byte>(_defragmentMap, Page.PageSize));
```
Scooletz, line 250: This is wrong I believe. You should always copy the payload first, otherwise you'll be overwriting the same memory. This is what I meant when discussing the `OperationsPerInvoke` so that you amortize the sets. The first step is to copy it to a working memory (a separate aligned page), then perform a few deletes & sets. My take would be to set at least 2 times or more so that the initial copy is not the major hit.

Scooletz (changes requested):

src/Paprika.Benchmarks/SlottedArrayBenchmarks.cs
```diff
@@ -191,6 +246,59 @@ public int EnumerateNibble(byte nibble)
...
+        // Insert additional key to trigger defragmentation
+        var additionalKey = Keccak.Zero;
+        Span<byte> value = stackalloc byte[32];
+
+        for (byte i = 0; i < 32; i++)
+        {
+            value[i] = i;
+        }
+
+        return map.TrySet(NibblePath.FromKey(additionalKey), value);
```
Scooletz, line 263: This performs just one set while the `CreateMap...` copies the page. Also, no keys are deleted. Is it a baseline?

src/Paprika.Benchmarks/SlottedArrayBenchmarks.cs
```diff
@@ -191,6 +246,59 @@ public int EnumerateNibble(byte nibble)
...
```
Scooletz, line 290: This sets just one value, right? What about the cost of the creating the map? Should `Defrag...` benchmark be used as a baseline?

src/Paprika.Tests/Data/SlottedArrayTests.cs
```diff
@@ -326,6 +326,90 @@ public void Defragmentation(int deleted)
...
+                .BeFalse("There should be no space in the map to make the defragmentation test work");
+
+            // Randomly select keys to be deleted.
+            HashSet<int> deletedKeys = new HashSet<int>(deleteCount);
+
+            for (var i = 0; i < deleteCount; i++)
+            {
+                var keyIndex = rand.Next(keyCount);
+
+                while (deletedKeys.Contains(keyIndex))
```
Scooletz, line 389: You can `while (deleted.Keys.Add(keyIndex) == false)`

src/Paprika/Data/SlottedArray.cs
```diff
@@ -904,80 +908,199 @@ private void MarkAsDeleted(int index)
...
+    /// </summary>
+    private uint CreateDeletedBitmask(ref ushort d, int offset)
+    {
+        uint deletedMask = 0;
+
+        // Extract the preamble data from the underlying slots and compare it with the preamble delete mask.
+        if (Vector256.IsHardwareAccelerated)
+        {
+            var slotData = Vector256.LoadUnsafe(ref d, (UIntPtr)offset);
+            var preambleData = Vector256.BitwiseAnd(slotData, Slot.GetKeyPreambleMaskAsVector256());
```
Scooletz, line 1041: This loads the data, then bitwise and it with the mask, then compares. Should we have it first checked if it's non-zero? I believe this is the approach taken by the `Span.IndexOf` as getting the significant bits where the value is zero is a bit more expensive. Again, to compare with the benchmarks

src/Paprika/Data/SlottedArray.cs
```diff
@@ -904,80 +908,199 @@ private void MarkAsDeleted(int index)
         hash = (ushort)~hash;
     }
 
+    /// <summary>
+    /// Defragments the underlying slotted array by overwriting the deleted slots. Performs vectorized defragmentation
+    /// by creating a deleted mask from the slot data and then moving only the alive slots, hashes and the corresponding
+    /// payload data.
+    /// </summary>
     private void Defragment()
```
Scooletz, line 916: `[SkipLocalsInit]` if not needed.

src/Paprika/Data/SlottedArray.cs
```diff
@@ -904,80 +908,199 @@ private void MarkAsDeleted(int index)
...
+                    setBitIndex = BitOperations.TrailingZeroCount(deletedMask);
+                    deletedMask ^= 1U << setBitIndex;
+                    alive--;
+                }
 
-                if (length > 0)
+                endIndex = setBitIndex + (i - HashesPerVector) / VectorsByBatch - 1;
+
+                // In case of consecutive deleted slots, find the next non-consecutive set bit.
+                while (startIndex == endIndex + 1 && deletedMask != 0)
```
Scooletz, line 977: Why this requires a loop? Can't it be that we find it bitwise without the loop?

src/Paprika.Benchmarks/SlottedArrayBenchmarks.cs
dipkakwani: One key is already deleted as part of the setup of this benchmark. This one directly triggers the defragment operation as part of TrySet. I believe this should serve as a good baseline for the `Large` benchmark. 

 `Large` benchmark also sets only one key, but it deletes multiple random keys based on the parameter. So in effect  `Large` one is capturing [key(s) deletion + defrag + one Set] time, while the smaller one is capturing [defrag + one Set] time.

src/Paprika/Data/SlottedArray.cs
dipkakwani: Nice catch! :grinning:  
Gave more thought and converted the loop into bit operations.

Scooletz (approved):
Some serious gains for smaller counts. Not worse for bigger. Let's merge!

PR #448, opened 2024-12-13
BufferPool stack capturing asserting mode

Scooletz:
This PR allows stack capturing assertion in the `BufferPool`. It's heavy so use only in tests and only when finding a leak.


dipkakwani:
Looks good. The previous approach was using` #if STACK_TRACE_TRACKING`, but this one is using enum for various levels. I hope there is no extra overhead with this approach for debug/release?

dipkakwani (approved):

Scooletz:
One check, we should be fine. But also, we can set it locally so should be better.

PR #449, opened 2024-12-16
Reduce RLP Compression & Decompression

dipkakwani:
Currently the memoised RLPs are stored in a compressed form when flushing to the `PageDb`, but when stored in-memory, these are in the decompressed form.

This PR makes RLPMemo's `Clear/Get` operation work on the compressed data, thus avoiding decompression. The `Set` operation still requires decompressed data since it may require extra allocation when operating with the compressed form. The current approach avoids any de-allocations/allocations with multiple `Clear/Set` operations as memory is handled by the caller. Also note that these changes optimize the `Compress` operation by making it possible to compress an already (partially) compressed RLPMemo, e.g. `Compress` -> `Clear(x)` ->`Clear(y)` -> `Compress` would make it optimally compressed without performing any de-allocations in the `Clear` operation.

Multiple approaches were considered to reduce the decompressions, including RLPMemo handling its own memory, allocating/de-allocating strategically while operating on the compressed data (double/half the size similar to std::Vector). But the current approach requires minimal changes/overhead to the existing code while providing a few optimizations.

### DIFF
--- a/src/Paprika.Tests/Chain/PrefetchingTests.cs
+++ b/src/Paprika.Tests/Chain/PrefetchingTests.cs
@@ -72,7 +72,7 @@ public class PrefetchingTests
     }
 
     [Test]
-    public async Task Makes_all_decompression_on_prefetch()
+    public async Task Makes_no_decompression_on_prefetch()
     {
         using var db = PagedDb.NativeMemoryDb(8 * 1024 * 1024, 2);
         var merkle = new ComputeMerkleBehavior(ComputeMerkleBehavior.ParallelismNone);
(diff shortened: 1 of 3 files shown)


Scooletz (changes requested):
I don't understand how it works and according to my understanding it should not work. It looks like some part of compression is still there (.Compress call) and that the structure does not respond to children resizing.

1. What is the size of RlpMemo? Let's consider a branch with nibbles set at `{0, 1, 2}`. It should be max 32 * 3 = `96` bytes long and potentially much less when some of the value are less than Keccak long and are not hashed for RLP.
2. How children removal is handled? Let's consider a branch with nibbles set at `{0, 1, 2}`. What happens with the `RlpMemo` when `2` is removed?
3. Why the call to `.Compress` still exists?

PR #455, opened 2024-12-19
Pooled span dicitionary direct pointer

### LINKED ISSUE #450: PooledSpanDictionary improvements
`PooledSpanDictionary` is the main component responsible for storing Paprika's data in-memory. It does by bucketing keys heavily using 16 pages as the initial fan out and then storing data as single linked lists. Currently, the addresses that are stored are 4 bytes long and point to a given page in the list to a given position (encoded as an `int`). `PooledSpanDictionary` is not thread safe as well, which requires providing `Child` commits when running `ComputeMerkleBehavior`. The following items could improve overall performance of the dictionary.

1. migrate address to a pointer
2. make `PooledSpanDictionary` concurrent for upserts/deletes

The 1st would introduce some overhead as currently address is 4 bytes long, when after migration would be 8 bytes long on x64. This would drastically reduce the lookup time though as the next item in a linked list could be accessed directly by following the pointer, rather then finding the page in the list and then performing the calculation. It should greatly reduce the read overhead over the `PooledSpanDictionary`. 

`PooledSpanDictionary` after #443 no longer requires visitor pattern to access the accounts and storage trees that were changed lately. This means, that as long as atomicity is preserved (both for allocation and for updates) we could make this dictionary much more like the `ConcurrentDictionary`. In regards to updating or finding the given key, a simple volatile at the top and walking down the linked list should be sufficient when we know that no colliding updates can happen. If this is not the case, an `Interlocked` lease could be used. It's worth to mention that the `ComputeMerkleBehavior` is designed in a way that there will be no colliding updates by multiple threads at the given time. This would mean, that updates in-situ, deletes and gets could be addressed using the technique described above. In regards to allocating new chunks, this could be done in a fully locked way, where the current working page is checked under a lock whether it can providing the required space, if not another would be added. As 1st point would ensure that we use direct pointers, no access to the list would be needed beside this allocation. This could allow to pass the same dictionary to all the work items and ditch the idea of the child commit (or just wrap the same dictionary)

Both should be measured extensively using benchmarks.

Scooletz:
This PR changes `PooledSpanDictionary` to use pointers rather than relative addresses that need to then transformed before accessing. This unfortunately makes the address size twice as big (from `uint` to `byte*`). When the `Root` `PageCount` is increased to make it big enough to keep the same number of buckets (`128kb` for the root instead of the earlier `64kb`), the new version wins.

### Benchmarks

`Read_write_small` has been modified.

#### Before pointer 

| Method                                  | Mean            | Error         | StdDev        | Gen0   | Code Size | Allocated |
|---------------------------------------- |----------------:|--------------:|--------------:|-------:|----------:|----------:|
| Read_from_big_dict                      | 174,569.2261 ns | 3,233.0694 ns | 3,024.2151 ns |      - |     410 B |         - |
| Enumerate_big_dict                      |  69,030.7471 ns |   670.5640 ns |   627.2460 ns |      - |     207 B |         - |
| Read_missing_with_hash_collisions       | 207,001.1198 ns | 2,432.6261 ns | 2,275.4799 ns |      - |     416 B |         - |
| Read_missing_with_no_hash_collisions    | 169,260.6755 ns | 1,962.7596 ns | 1,835.9666 ns |      - |     435 B |         - |
| Read_write_small                        |   3,351.7085 ns |    22.0201 ns |    18.3878 ns |   1,364 B |         - |
| Read_colliding_hashes_different_lengths |       0.8284 ns |     0.0102 ns |     0.0095 ns |      - |     660 B |         - |

#### After pointer and 2 times bigger Root

| Method                                  | Mean            | Error         | StdDev        | Gen0   | Code Size | Allocated |
|---------------------------------------- |----------------:|--------------:|--------------:|-------:|----------:|----------:|
| Read_from_big_dict                      | 143,440.3209 ns | 2,785.0180 ns | 3,420.2525 ns |      - |     343 B |         - |
| Enumerate_big_dict                      |  41,784.9841 ns |   742.0425 ns |   694.1071 ns |      - |     133 B |         - |
| Read_missing_with_hash_collisions       | 162,168.6432 ns | 3,115.8957 ns | 3,060.2262 ns |      - |     349 B |         - |
| Read_missing_with_no_hash_collisions    | 126,322.7669 ns | 2,015.2744 ns | 1,885.0889 ns |      - |     381 B |         - |
| Read_write_small                        |   3,009.3006 ns |    35.7414 ns |    33.4325 ns |   1,301 B |         - |
| Read_colliding_hashes_different_lengths |       0.7309 ns |     0.0155 ns |     0.0145 ns |      - |     590 B |         - |


src/Paprika/Chain/PooledSpanDictionary.cs
```diff
@@ -43,9 +50,15 @@ public PooledSpanDictionary(BufferPool pool, bool preserveOldValues = false)
         var pages = new Page[Root.PageCount];
         for (var i = 0; i < Root.PageCount; i++)
         {
-            pages[i] = RentNewPage(true);
+            pages[i] = RentNewPage(false);
```
dipkakwani, line 53: Interesting use of clearing the page in parallel. Makes me wonder if we could make a bulk/batch rent API as well, instead of sequentially renting a new page? Along the same lines, would exposing a batch Set API in `PooledSpanDictionary` could also be useful?

PR #456, opened 2024-12-26
Remove Compression & Decompression from RLPMemo

dipkakwani:
Currently the memoised RLPs are stored in a compressed form when flushing to the PageDb, but when stored in-memory, these are in the decompressed form.

This PR removes compression & decompression altogether by always storing the memorized RLPs in compressed form. To achieve this, a NibbleSet index (2 bytes) is stored at the end of RLPMemo buffer which indicates whether each element is currently stored in the memo. For example, `[k1 | k2 | k5]` in RLPMemo would also store NibbleSet `{1, 2, 5}` or `0b0000 0000 0010 0110` to show that only 1, 2 & 5 index children's keccaks are currently memoized.

Also note that due to an existing optimization done during DB write operation to not store memoized RLPs for top level state branches (`SkipRlpMemoizationForTopLevelsCount`), the RLPMemo data structure supports storing empty buffer even if there are children for that branch.

Metrics with importing mainnet data  (with parallelization enabled along with a few optimizations):

image

`Root: 0xe64f828043054f23455c65871e737b3f03c98599f7feb5bea4f450c7ca089dd4 was being imported to Paprika in 4:54:04.5507713 and resulted in 0xe64f828043054f23455c65871e737b3f03c98599f7feb5bea4f450c7ca089dd4`

For comparison the baseline `importer` branch was also used for a partial import (~70GB DB size, NOT a full import)

image

Most of the metrics look similar except for the buffer pool size and commit time, which is better with the rlpmemo changes. 
Overall, the changes simplify the way memoization is done by removing the overhead of compression & decompression from various places (write to DB, prefetch etc.).

### DIFF
--- a/src/Paprika.Tests/Merkle/RlpMemoTests.cs
+++ b/src/Paprika.Tests/Merkle/RlpMemoTests.cs
@@ -200,7 +200,7 @@ public class RlpMemoTests
         var children = new NibbleSet();
 
         Span<byte> keccak = new byte[Keccak.Size];
-        keccak.Fill(0xFF);            
+        keccak.Fill(0xFF);
 
         for (var i = 0; i < RlpMemo.MaxSize - NibbleSet.MaxByteSize; i++)
         {
--- a/src/Paprika.Tests/Merkle/RlpMemoTests.cs
+++ b/src/Paprika.Tests/Merkle/RlpMemoTests.cs
@@ -1,15 +1,11 @@
 using FluentAssertions;
-using Paprika.Chain;
 using Paprika.Crypto;
-using Paprika.Data;
 using Paprika.Merkle;
-using Paprika.RLP;
 
 namespace Paprika.Tests.Merkle;
 
 public class RlpMemoTests
 {
-
     // All the write operations on RlpMemo
     private enum RlpMemoOperation
     {
@@ -22,16 +18,11 @@ public class RlpMemoTests
     [Test]
     public void Insert_get_operation()
     {
-        Span<byte> raw = [];
         Span<byte> workingMemory = new byte[RlpMemo.MaxSize];
-
-        Span<byte> keccak = new byte[Keccak.Size];
-        keccak.Fill(0xFF);
-
         NibbleSet.Readonly children = new NibbleSet(0xA, 0xB, 0xC);
+        var memo = new RlpMemo([]);
 
-        var memo = new RlpMemo(raw);
-        memo = InsertKeccak(memo, children, keccak, workingMemory);
+        InsertRandomKeccak(ref memo, children, out var data, workingMemory);
 
         memo.Length.Should().Be(GetExpectedSize(children.SetCount));
 
@@ -41,33 +32,33 @@ public class RlpMemoTests
             {
                 memo.Exists(i).Should().BeTrue();
                 memo.TryGetKeccak(i, out var k).Should().BeTrue();
-                k.SequenceEqual(keccak).Should().BeTrue();
+                k.SequenceEqual(data[i].Span).Should().BeTrue();
             }
         }
+
+        CompareMemoAndDict(memo, data);
     }
 
     [Test]
     public void Set_get_operation()
     {
-        Span<byte> raw = [];
         Span<byte> workingMemory = new byte[RlpMemo.MaxSize];
-
-        Span<byte> keccak = new byte[Keccak.Size];
-        keccak.Fill(0xFF);
-
-        Span<byte> keccakNew = new byte[Keccak.Size];
-        keccakNew.Fill(0xAA);
-
         NibbleSet.Readonly children = new NibbleSet(0xA, 0xB, 0xC);
+        var memo = new RlpMemo([]);
 
-        var memo = new RlpMemo(raw);
-        memo = InsertKeccak(memo, children, keccak, workingMemory);
+        InsertRandomKeccak(ref memo, children, out var data, workingMemory);
+
+        memo.Length.Should().Be(GetExpectedSize(children.SetCount));
+        
+        var random = new Random(13);
 
         for (byte i = 0; i < NibbleSet.NibbleCount; i++)
         {
             if (children[i])
             {
-                memo.Set(keccakNew, i);
+                data[i] = random.NextKeccak();
+                memo.Set(data[i].Span, i);
+                CompareMemoAndDict(memo, data);
             }
         }
 
@@ -79,30 +70,31 @@ public class RlpMemoTests
             {
                 memo.Exists(i).Should().BeTrue();
                 memo.TryGetKeccak(i, out var k).Should().BeTrue();
-                k.SequenceEqual(keccakNew).Should().BeTrue();
+                k.SequenceEqual(data[i].Span).Should().BeTrue();
             }
         }
+
+        CompareMemoAndDict(memo, data);
     }
 
     [Test]
     public void Clear_get_operation()
     {
-        Span<byte> raw = [];
         Span<byte> workingMemory = new byte[RlpMemo.MaxSize];
-
-        Span<byte> keccak = new byte[Keccak.Size];
-        keccak.Fill(0xFF);
-
         NibbleSet.Readonly children = new NibbleSet(0xA, 0xB, 0xC);
+        var memo = new RlpMemo([]);
 
-        var memo = new RlpMemo(raw);
-        memo = InsertKeccak(memo, children, keccak, workingMemory);
+        InsertRandomKeccak(ref memo, children, out var data, workingMemory);
+
+        memo.Length.Should().Be(GetExpectedSize(children.SetCount));
 
         for (byte i = 0; i < NibbleSet.NibbleCount; i++)
         {
             if (children[i])
             {
+                data[i] = Keccak.Zero;
                 memo.Clear(i);
+                CompareMemoAndDict(memo, data);
             }
         }
 
@@ -117,27 +109,28 @@ public class RlpMemoTests
                 k.IsEmpty.Should().BeTrue();
             }
         }
+
+        CompareMemoAndDict(memo, data);
     }
 
     [Test]
     public void Delete_get_operation()
     {
-        Span<byte> raw = [];
         Span<byte> workingMemory = new byte[RlpMemo.MaxSize];
-
-        Span<byte> keccak = new byte[Keccak.Size];
-        keccak.Fill(0xFF);
-
         NibbleSet.Readonly children = new NibbleSet(0xA, 0xB, 0xC);
+        var memo = new RlpMemo([]);
 
-        var memo = new RlpMemo(raw);
-        memo = InsertKeccak(memo, children, keccak, workingMemory);
+        InsertRandomKeccak(ref memo, children, out var data, workingMemory);
+
+        memo.Length.Should().Be(GetExpectedSize(children.SetCount));
 
         for (byte i = 0; i < NibbleSet.NibbleCount; i++)
         {
             if (children[i])
             {
+                data.Remove(i);
                 memo = RlpMemo.Delete(memo, i, workingMemory);
+                CompareMemoAndDict(memo, data);
             }
         }
 
@@ -152,6 +145,8 @@ public class RlpMemoTests
                 k.IsEmpty.Should().BeTrue();
             }
         }
+
+        CompareMemoAndDict(memo, data);
     }
 
     [Test]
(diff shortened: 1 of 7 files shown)


src/Paprika/Merkle/RlpMemo.cs
```diff
@@ -64,139 +77,118 @@ public bool TryGetKeccak(byte nibble, out ReadOnlySpan<byte> keccak)
...
-        if (leftover.IsEmpty)
-        {
-            // no RLP cached yet
-            span.Clear();
-            return new RlpMemo(span);
-        }
-
-        if (leftover.Length == Size)
+        // Check if the element exists
+        if (_buffer.Length == 0 || ((ushort)children & (1U << nibble)) == 0)
```
Scooletz, line 83: Does it effectively check whether the child is set in children? If yes, isn't it `RlpMemo` agnostic and `NibbleSet.Readonly[nibble]` could be used directly in the call site? Then, have this method removed?

src/Paprika/Merkle/RlpMemo.cs
```diff
@@ -64,139 +77,118 @@ public bool TryGetKeccak(byte nibble, out ReadOnlySpan<byte> keccak)
...
+        // Compute the destination size for copying.
+        var size = children.SetCount * Keccak.Size;
+
+        Debug.Assert(workingSet.Length >= size);
+        var span = workingSet[..size];
 
-        var memo = new RlpMemo(ComputeMerkleBehavior.MakeRlpWritable(memoizedRlp));
-        var at = 0;
+        // Ensure that this element already exists in the list of children
+        var leftChildren = (ushort)((ushort)children & ((1U << (nibble + 1)) - 1));
```
Scooletz, line 122: Use counters in the children maybe? `SetCountBefore` or whatever it was named?

src/Paprika/Merkle/ComputeMerkleBehavior.cs
```diff
@@ -546,16 +543,24 @@ private void EncodeBranch(scoped in Key key, scoped in ComputeContext ctx, scope
...
                         memoizedUpdated = true;
-                        memo.Set(keccakOrRlp, i);
+
+                        if (memo.Exists(i, branch.Children))
+                        {
+                            memo.Set(keccakOrRlp, i, branch.Children);
+                        }
+                        else if (keccakOrRlp.DataType == KeccakOrRlp.Type.Keccak)
+                        {
+                            memo = RlpMemo.Insert(memo, i, branch.Children, keccakOrRlp.Span, rlpMemoization);
```
Scooletz, line 553: What happens if encoded data in are not a `Keccak`? Do we store a `Keccak.Zero` then? If not, how does it distinguish between existence of a keccak or not? Doesn't it mean that RlpMemo requires a similar NibbleSet to store bits of what was remembered or not?

Scooletz (changes requested):
I think I either don't follow or there's a mistake in the approach of removing the `Compress`.

Let's assume that we have a `Branch` that has the children with the following first nibbles `{A, B, C}`. How does this approach distinguish cases where:

1. `A` and `B` are encoded to `Keccak` while `C` RLP is less than 31 bytes
2  `A` and `C` are encoded to `Keccak` while `B` RLP is less than 31 bytes

Where is the information stored now?

src/Paprika/Merkle/RlpMemo.cs
```diff
@@ -64,139 +57,147 @@ public bool TryGetKeccak(byte nibble, out ReadOnlySpan<byte> keccak)
...
-        // It's neither empty nor full. It must be the compressed form, prepare setup first
-        span.Clear();
-        var memo = new RlpMemo(span);
+        var nibbleIndex = index.SetCountBefore(nibble) - 1;
+        return _buffer.Slice(nibbleIndex * Keccak.Size, Keccak.Size);
+    }
 
-        // Extract empty bits if any
-        NibbleSet.Readonly empty;
+    private void GetIndex(out NibbleSet.Readonly index)
```
Scooletz, line 91: The `NibbleSet.Readonly` is only 2 bytes long. Have you considered returning it by value?

src/Paprika/Merkle/RlpMemo.cs
```diff
@@ -64,139 +57,147 @@ public bool TryGetKeccak(byte nibble, out ReadOnlySpan<byte> keccak)
...
+        var indexLength = _buffer.Length % Keccak.Size;
 
-        // The empty bytes length is anything that is not aligned to the Keccak size
-        var emptyBytesLength = leftover.Length % Keccak.Size;
-        if (emptyBytesLength > 0)
+        if (indexLength != 0)
         {
-            var bits = leftover[^emptyBytesLength..];
-            NibbleSet.Readonly.ReadFrom(bits, out empty);
+            var bits = _buffer[^indexLength..];
```
Scooletz, line 98: This might be a nitpicking, but if we stored the index at the beginning, no span slice would be required and we could get it from the start, just by passing the buffer.

src/Paprika/Merkle/RlpMemo.cs
```diff
@@ -64,139 +57,147 @@ public bool TryGetKeccak(byte nibble, out ReadOnlySpan<byte> keccak)
...
+            NibbleSet.Readonly.ReadFrom(bits, out index);
         }
         else
         {
-            empty = default;
-        }
+            Debug.Assert(_buffer.Length is 0 or MaxSize);
```
Scooletz, line 103:
```suggestion
            Debug.Assert(_buffer.Length is 0 or MaxSize, "Only empty or full RlpMemo can have no index");
```

src/Paprika/Merkle/RlpMemo.cs
```diff
@@ -64,139 +57,147 @@ public bool TryGetKeccak(byte nibble, out ReadOnlySpan<byte> keccak)
...
-        if (children.SetCount == 2 && (key.Path.Length + key.StoragePath.Length) % 2 == 0)
-        {
-            return 0;
-        }
+        memo.GetIndex(out var index);
 
-        var memo = new RlpMemo(ComputeMerkleBehavior.MakeRlpWritable(memoizedRlp));
-        var at = 0;
+        // Ensure that this element doesn't already exist.
+        Debug.Assert(!index[nibble]);
```
Scooletz, line 121: Turn comment into the debug message maybe?

src/Paprika/Merkle/RlpMemo.cs
```diff
@@ -64,139 +57,147 @@ public bool TryGetKeccak(byte nibble, out ReadOnlySpan<byte> keccak)
...
 
-        var empty = new NibbleSet();
+        // Compute the destination size for copying.
+        var size = (index.SetCount < NibbleSet.NibbleCount - 1)
```
Scooletz, line 124: Size computation is repeated into multiple places. Why don't we extract it as a method? It'd be based on the number of destination nibbles set right? So a simple static method would do?

src/Paprika/Merkle/RlpMemo.cs
```diff
@@ -64,139 +57,147 @@ public bool TryGetKeccak(byte nibble, out ReadOnlySpan<byte> keccak)
...
 
-    private static volatile bool _decompressionForbidden;
+        // Ensure that this element isn't already deleted.
+        Debug.Assert(index[nibble]);
 
-    public readonly struct NoDecompressionScope : IDisposable
-    {
-        public NoDecompressionScope()
+        // Compute the destination size for copying.
+        var size = (index.SetCount < NibbleSet.NibbleCount)
```
Scooletz, line 168: Again, use the extracted method?

src/Paprika/Merkle/ComputeMerkleBehavior.cs
```diff
@@ -729,15 +714,22 @@ private void EncodeBranch(scoped in Key key, scoped in ComputeContext ctx, scope
                 stream.Encode(value);
                 if (memoize)
                 {
+                    memoizedUpdated = true;
+
                     if (value.Length == Keccak.Size)
                     {
-                        memoizedUpdated = true;
-                        memo.SetRaw(value, i);
+                        if (memo.Exists(i))
```
Scooletz, line 721: I like the clear approach here, where different cases are split.

src/Paprika/Merkle/RlpMemo.cs
```diff
@@ -64,139 +57,147 @@ public bool TryGetKeccak(byte nibble, out ReadOnlySpan<byte> keccak)
...
     }
 
-    [SkipLocalsInit]
-    public static int Compress(in Key key, scoped in ReadOnlySpan<byte> memoizedRlp, NibbleSet.Readonly children, scoped in Span<byte> writeTo)
+    public static RlpMemo Insert(RlpMemo memo, byte nibble, ReadOnlySpan<byte> keccak, scoped in Span<byte> workingSet)
     {
-        // Optimization, omitting some of the branches to memoize.
-        // It omits only these with two children where the cost of the recompute is not big.
-        // To prevent an attack of spawning multiple levels of such branches, only even are skipped
-        if (children.SetCount == 2 && (key.Path.Length + key.StoragePath.Length) % 2 == 0)
```
Scooletz, line 136: This shaved off some memory for small branches. It wasn't the biggest saving but it did something. Is it intentionally removed?

src/Paprika/Merkle/RlpMemo.cs
```diff
@@ -64,139 +57,147 @@ public bool TryGetKeccak(byte nibble, out ReadOnlySpan<byte> keccak)
...
+    {
+        // Extract the index bits.
+        var indexLength = _buffer.Length % Keccak.Size;
 
-        // The empty bytes length is anything that is not aligned to the Keccak size
-        var emptyBytesLength = leftover.Length % Keccak.Size;
-        if (emptyBytesLength > 0)
+        if (indexLength != 0)
         {
-            var bits = leftover[^emptyBytesLength..];
```
Scooletz, line 103: We loose one compression opportunity here. Imagine a branch that has only one Keccak memoized in RlpMemo. In such a cache, we could encode this one on a single byte. Even more, we could encode up to 2 on a single byte. This would, for cases where only 1 or 2 keccaks are memoized, make them ligher by one byte. For 2 keccaks, it would be ~ 1.5% saving where for one it would be 3%. This would make it a bit more complicated though. Another option would be to treat the special case of 1 byte to keep the not-set (which would save some space for 15 and 14 set, but would be much less impactful size-wise I think).

I know that original decompress/compress had no such feature https://github.com/NethermindEth/Paprika/blob/61c326a39e942ab9c550f6a6b9b5df59e67c0f0f/src/Paprika/Merkle/RlpMemo.cs#L171-L176 but we use something similar for branches themselves https://github.com/NethermindEth/Paprika/blob/61c326a39e942ab9c550f6a6b9b5df59e67c0f0f/src/Paprika/Merkle/Node.cs#L405-L416

src/Paprika.Tests/Merkle/RlpMemoTests.cs
```diff
@@ -1,139 +1,276 @@
...
-        Run(raw, 0, NibbleSet.Readonly.All, OddKey);
+        Set,
+        Clear,
+        Delete,
+        Insert
     }
 
     [Test]
-    public void All_children_set_with_all_Keccaks_set()
+    public void Random_delete()
```
Scooletz, line 22: The random tests are nice for fuzzying, but it would be beneficial to have some explicit A, B, C where A, B or A,C are set so that we cover what previously was missing. Or maybe I'm not seeing the tests like this but they do exist 👀

Scooletz (changes requested):
A few remarks, but almost ready to merge.

src/Paprika.Tests/Merkle/RlpMemoTests.cs
dipkakwani: Yes, that test is missing but I am not quite sure how to capture this in the test - a child with Rlp < 31 bytes while others have keccaks?
I looked at existing tests to figure it out (e.g. `Three_accounts_sharing_start_nibble`, `Branch_two_leafs`), but I couldn't find a way to simulate this scenario:

```
[Test]
public void Insert_get_operation()
{
        var commit = new Commit();

        const string key0 = "A0000001";
        const string key1 = "B0000002";
        const string key2 = "C0000003";
        
        commit.Set(key0);
        commit.Set(key1);
        commit.Set(key2);
        
        NibbleSet.Readonly children = new NibbleSet(0xA, 0xB, 0xC);
        commit.SetBranch(Key.Merkle(NibblePath.Empty), children);

        var merkle = new ComputeMerkleBehavior();

        merkle.BeforeCommit(commit, CacheBudget.Options.None.Build());
}
```
I am probably missing some obvious way to complete this test scenario :face_with_diagonal_mouth:

src/Paprika/Merkle/RlpMemo.cs
dipkakwani: That's a good point! I was also thinking of another approach and wanted to run through you to check which one might be better (before completing the code changes): 
To maximum utilize one byte, it could store partial bitset index. We can basically divide the 2 bytes into two parts and utilize only 1 byte to represent the index for cases where multiple contiguous indices are set.

Original (2 bytes) : [0, 1, 2, .. 15]
Modified (1 Byte): [x, 0, 1, 2, 3, 4, 5, 6] or [x, 7, 8, 9, 10, 11, 12, 13]

x can be 0 or 1 depending on which half it is representing.

As long as many children fall in a contiguous range, we would be only needing 1 byte. Disadvantage to this approach :
a) even if 2 children exist in two different halves, we would have to fall back on 2 bytes of storage. 
b) 14,15 children can never be represented with one byte

Does this make sense? Do you think it is better/worse than your proposed approach?

Scooletz: What an interesting idea! One would need to count the collisions and probabilities 😅 

- for 2 nibbles, picking up two from one of the ranges is 7/15. 
- for 3 nibbles, we get 1/5
- for 4 nibbles, we get 1/13
- 
As we're trying to optimize for the small branches (the big won't benefit from removing 1 byte) it's a bet. With the ncoding proposed before, we get 1 byte for nibbles and always 2 bytes for other cases. Here we optimize up to any number, but this is diminishing returns, right? With 4 Keccaks it'll be only 1% saved.

What do you think then?

src/Paprika.Tests/Merkle/RlpMemoTests.cs
Scooletz: I think I'd replace the Branch based test with a direct `KeccakMemo` test maybe? Not sure if covering the compounded Branch + memo is doable in a nice way.

src/Paprika/Merkle/RlpMemo.cs
dipkakwani: Good catch, I moved this code to `InspectBeforeApply`, since this optimization was done during compression, which we don't do now. But note that for this particular check, we now have to parse the branch data in `InspectBeforeApply` and then find out the `children.SetCount` value. Is the extra parsing in `InspectBeforeApply` still worth it to save off for small branches?

src/Paprika/Merkle/RlpMemo.cs
dipkakwani: I would like to take up this optimization in another PR, since it would require extra logic and even additional testing to ensure that it works fine.

src/Paprika/Merkle/RlpMemo.cs
dipkakwani: I tried this particular change and for some reason it started breaking the `CalculateStateRootHash`. Tried bunch of ways to test/debug, but it wasn't clear what exactly was causing the test failure. Moving back to the original index at the end resolved the issues. 
Do you have any pointers if that is expected? But either way, I prefer to keep the underlying data consistent with the original representation after `Compress` operation, by keeping the index at the end and not deal with state root mismatch issues :sweat:

src/Paprika.Tests/Merkle/RlpMemoTests.cs
dipkakwani: Added basic new tests in RLPMemo with fixed children. Branch + memo test couldn't be done without exposing few internal details as public, so sticking to independent unit tests.

src/Paprika/Merkle/RlpMemo.cs
dipkakwani: As discussed offline, added new tests with grow/shrink which were able to catch the issue with header placement/incorrect implementation. Also modified the `Large_random_operations` test which caught failures too. The current coverage for RLPMemo class is 100%. All the test pass with the current implementation.

src/Paprika/Merkle/RlpMemo.cs
Scooletz: Having this said, should this PR be no longer labeled with `Breaking`?

src/Paprika/Merkle/RlpMemo.cs
Scooletz: Massive.

Scooletz (commented):
A few more clarifications.

src/Paprika/Merkle/RlpMemo.cs
dipkakwani: Yes, removed the label.

src/Paprika/Merkle/RlpMemo.cs
dipkakwani: I didn't get you fully.. we can save up that space, but I just wanted to point out that here we are doing an additional bit of parsing to find out the children count. I think that it is fine to be done for every branch, since we would be saving 64+2 bytes potentially with the smaller branches.

Scooletz: We are. This is due to the nature of unmaterialized data that need to be reparsed on each passing. I cannot provide you with numbers (#286 has none), but saving 66 bytes out of 68 for small branches seem to be a good deal 👀

Scooletz (approved):
Massive. The pool size speaks for itself. Let's merge in.

Issue #461, opened 2025-01-22
RLPMemo Index Storage Optimizations


dipkakwani:
The RLPMemo currently uses the following structure:
[k1 | k2 | ... | kn | index]

where k1, k2...kn are the memoized keccaks and index is a bitmask indicating which nibbles out of `NibbleSet.NibbleCount` are memoized in the memo. The index currently uses a fixed size 2 bytes (`ushort`) and it is stored at the end of the RLPMemo. Two optimizations are possible with respect to the index:
1. Store the index at the start instead of the end to avoid frequent slicing of RLPMemo's buffer.  For example, if the index was stored at the start, the frequently used `Exist` operation would become faster.
2. Use 1 byte index when there are <= 2 keccaks in the RLPMemo. It is possible to directly store the index of up to 2 elements instead of the bitmask representation to save up 1 extra byte in the index. For example, if there were only 2 elements in RLPMemo at 4 & 15, both these indices could be directly encoded in a single byte. Ideally, we should modify `NibbleSet` so that it's capable of writing/reading from/to 1-2 bytes instead of fixed 2 bytes only.

PR #463, opened 2025-01-27
RLPMemo Move Index (Index Optimization Part 1)

dipkakwani:
Move the index to start in RLPMemo as described in https://github.com/NethermindEth/Paprika/issues/461
Also adds a new test for verifying the expected layout of the RLPMemo data structure.

Test binary diff:

image
(Index = 0x 001C, moved from end to start)

Perf metrics:

image

`Root: 0xe64f828043054f23455c65871e737b3f03c98599f7feb5bea4f450c7ca089dd4 was being imported to Paprika in 4:33:22.8327645 and resulted in 0xe64f828043054f23455c65871e737b3f03c98599f7feb5bea4f450c7ca089dd4`

Previous metrics for reference (without this change):
image


src/Paprika.Tests/Merkle/RlpMemoTests.cs
```diff
@@ -430,6 +430,20 @@ public void Large_random_operations(int numOperations)
...
+    {
+        Span<byte> workingMemory = new byte[RlpMemo.MaxSize];
+        NibbleSet.Readonly children = new NibbleSet(0xA, 0xB, 0xC);
+        var memo = new RlpMemo([]);
+
+        InsertRandomKeccak(ref memo, children, out _, workingMemory);
+
+        // Expected layout:
+        // [index (0b 0000 0000 0001 1100) | keccakA | keccakB | keccakC]
+        return Verify(workingMemory.ToArray());
```
Scooletz, line 444: Should it not be that the output is only verified, not the whole working memory? I think we should be fine with the rest of memory being messed up. Unless you want to ensure that the rest is kept intact?

Scooletz (commented):
A minor on the `Verify`. The rest looks awesome and once we clarify we can merge as soon as possible.

src/Paprika.Tests/Merkle/RlpMemoTests.cs
dipkakwani: Fixed by verifying only the memo's buffer.

Scooletz (approved):

PR #472, opened 2025-02-10
New BottomPage

Scooletz:
A brand new structure of the `BottomPage` that handles data distribution in a better way. 

First, when the bottom with no children is full, one child is allocated. This makes the bottom page half-full on average.

When no more space, the main map is tried to be flushed down to existing ones. If it fails a new one is allocated and data are redistributed to satisfy `nibble0 >= child_index` assertion. Effectively, if a child is inserted in the middle (it will always have as the first one is allocated at 0), it will steal data from its left neighbor. Then the main map is tried to be redistributed again.

When there's no more option to make progress, the bottom page is transformed into a `DataPage`. This is done without copying the main page as it has the same payload structure as `DataPage`. The only thing that is accessed is the child redistribution.

### Benchmarks

The Ethereum `mainnet` database has been shrunk from `275GB` to `237GB` which gives ~14% reduction in size

#### Import

image

#### CLI stats

CLI `stats` show reduction in tree depth. The state is now 5 levels deep (beside the root that allows to jump first 2 nibbles). The storage with 3 levels of fan out at the top (the first is in the root), followed by 7 levels of the storage. This means that for a single storage read max 7 SlottedArrays will be scanned.

```ps
Paprika\src\Paprika.Cli > dotnet run -c Release -- stats E:\db\ 320 64
```

image

### DIFF
--- a/src/Paprika.Tests/Store/AbandonedTests.cs
+++ b/src/Paprika.Tests/Store/AbandonedTests.cs
@@ -84,10 +84,10 @@ public class AbandonedTests : BasePageTests
 
     [TestCase(20, 1, 10_000, false, TestName = "Accounts - 1")]
     [TestCase(428, 100, 10_000, false, TestName = "Accounts - 100")]
-    [TestCase(21796, 4000, 200, false,
+    [TestCase(20067, 4000, 200, false,
         TestName = "Accounts - 4000 to get a bit reuse",
         Category = Categories.LongRunning)]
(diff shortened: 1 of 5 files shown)


Scooletz:
Could it be that the top bucket is optimized so that it's capable of storing full merkle 512 + 1 byte of the `Node.Branch` plus a fanout of 16 for trees that are big enough?

PR #474, opened 2025-02-18
Side-car Merkle

Scooletz:
This PR builds on top of #472 by adding a side-car page to keep Merkle related data. This allows keeping the Merkle off the main `DataPage` map, leaving it as a whole for the write-through caching purposes. The side-car itself uses the new scaling design of the `BottomPage`. 

This PR does not change neither the database size nor the number of pages used per commit when running `Importer`. 

The depth and the shape of the tree is yet to be searched for using `Cli stats`. In the current, non-prefetched (see: #396) way of scanning it takes much too long.


Scooletz:
Implemented under #477

PR #482, opened 2025-05-22
ContractId introduced

Scooletz:
This PR introduces a notion of `ContractId`. This is a simple high level wrap around an `uint` increasing readability of the codebase.


asdacap (approved):
Much better. Thank you.

